PR #39, opened 2024-02-21
feat: pull the Connect server's version (lazily)

nealrichardson:
Just a quick PR to add a little feature we will need, but also to demonstrate how to use `responses` to stand in for the Connect API. Very straightforward to add a feature that makes an API request and then set up a test that uses a simulated response.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,7 @@ build==1.0.3
 coverage==7.4.0
 mypy==1.8.0
 pytest==7.4.4
+responses>=0.25
 ruff==0.1.14
 setuptools==69.0.3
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -2,6 +2,7 @@ build==1.0.3
 coverage==7.4.0
 mypy==1.8.0
 pytest==7.4.4
+responses>=0.25
```
tdstein, line 5:
```suggestion
responses==0.25
```

nit; I've pinned other dependencies to avoid unexpected changes to behavior.

tdstein (approved):
LGTM

PR #41, opened 2024-02-21
build: disables coverage threshold for new files

tdstein:


.github/workflows/pull-request.yaml
```diff
@@ -43,6 +43,4 @@ jobs:
       with:
           coverageFile: coverage.xml
           thresholdAll: 0.8
-          thresholdNew: 0.9
```
nealrichardson, line 46: Why, out of curiosity?

tdstein: It was causing CI to fail on files that were updated during refactoring. Similar to the behavior that drove the change to set `thresholdAll: 0.8`. I'd like to re-enable it soon once things have settled.

PR #45, opened 2024-02-21
fix: allow chaining paths with urls.append_path

tdstein:
This change moves base URL validation to the Config initialization sequence, which allows `append_path` to be called on URLs that are not base URLs.

### DIFF
--- a/src/posit/connect/urls.py
+++ b/src/posit/connect/urls.py
@@ -5,7 +5,7 @@ import posixpath
 from urllib.parse import urlsplit, urlunsplit
 
 
-def fix(url: str) -> str:
+def server_to_api_url(url: str) -> str:
     """
     Fixes the given URL by appending '__api__' if it doesn't already end with it.
 
@@ -15,6 +15,7 @@ def fix(url: str) -> str:
     Returns:
         str: The fixed URL.
     """
+    url = url.rstrip("/")
     if not url.endswith("__api__"):
         return append_path(url, "__api__")
     return url
@@ -31,7 +32,7 @@ def validate(url: str) -> None:
         bool: True if the URL is valid, False otherwise.
 
     Raises:
-        ValueError: If the URL is missing a scheme, is not absolute, or does not end with '/__api__'.
+        ValueError: If the URL is missing a scheme or is not absolute.
     """
     split = urlsplit(url, allow_fragments=False)
     if not split.scheme:
@@ -42,11 +43,6 @@ def validate(url: str) -> None:
     if not split.netloc:
         raise ValueError(
             f"url must be absolute (e.g., http://example.com/__api__): {url}"
-        )
-
-    if not (split.path and split.path.endswith("/__api__")):
-        raise ValueError(
-            f"url must end with path __api__ (e.g., http://example.com/__api__): {url}"
         )
 
 
--- a/src/posit/connect/urls.py
+++ b/src/posit/connect/urls.py
@@ -5,23 +5,33 @@ import posixpath
 from urllib.parse import urlsplit, urlunsplit
 
 
-def append_path(url: str, path: str) -> str:
+def fix(url: str) -> str:
     """
-    Appends a path to the end of a URL.
+    Fixes the given URL by appending '__api__' if it doesn't already end with it.
 
     Args:
-        url (str): The URL to append the path to.
-        path (str): The path to append to the URL.
+        url (str): The URL to fix.
 
     Returns:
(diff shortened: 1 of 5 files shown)


src/posit/connect/urls_test.py
```diff
@@ -1,28 +1,31 @@
...
     with pytest.raises(ValueError):
-        append_path("http://foo.bar/baz", "qux")
+        validate("http://foo.bar/baz")
 
 
-def test_fix_with_no_errors():
-    url = append_path("http://foo.bar/baz/__api__", "qux")
-    assert url == "http://foo.bar/baz/__api__/qux"
+def test_append_path():
+    append_path("http://foo.bar/__api__", "baz") == "http://foo.bar/__api__/baz"
```
nealrichardson, line 27:
```suggestion
    assert append_path("http://foo.bar/__api__", "baz") == "http://foo.bar/__api__/baz"
```

src/posit/connect/urls_test.py
```diff
@@ -1,28 +1,31 @@
...
+
+
+def test_append_path_with_leading_slash():
+    append_path("http://foo.bar/__api__", "/baz") == "http://foo.bar/__api__/baz"
```
nealrichardson, line 31:
```suggestion
    assert append_path("http://foo.bar/__api__", "/baz") == "http://foo.bar/__api__/baz"
```

src/posit/connect/urls.py
```diff
@@ -5,23 +5,18 @@
...
-        str: The modified URL with the appended path.
+        bool: True if the URL is valid, False otherwise.
 
     Raises:
-        ValueError: If the URL does not specify a scheme, is not absolute, or does not end with "/__api__".
-
-    Example:
-        >>> append_path("http://example.com", "api")
-        'http://example.com/__api__/api'
+        ValueError: If the URL is missing a scheme, is not absolute, or does not end with '/__api__'.
```
nealrichardson, line 19: I don't think we should require users to add `__api__` to their server URLs. Convention everywhere in our [docs](https://docs.posit.co/connect/cookbook/), and presumably in the env vars we inject into content, is that `__api__` is not included in the URL you provide:

> Your Connect server URL is the same URL you use to access the Connect dashboard, minus the connect path. If you access the dashboard at https://rsc.company.com/connect/, the server URL is https://rsc.company.com/.

We should add it if it is not present, IMO.

src/posit/connect/urls_test.py
tdstein: Ha, good catch.

src/posit/connect/urls.py
tdstein: Good call. I'll add that back in.

src/posit/connect/urls.py
```diff
@@ -5,23 +5,34 @@
 from urllib.parse import urlsplit, urlunsplit
 
 
-def append_path(url: str, path: str) -> str:
+def server_to_api_url(url: str) -> str:
```
nealrichardson, line 8: @tdstein I renamed the function, `fix()` seemed vague, LMK what you think.

src/posit/connect/urls.py
```diff
@@ -34,12 +45,23 @@ def append_path(url: str, path: str) -> str:
             f"url must be absolute (e.g., http://example.com/__api__): {url}"
         )
 
-    if not (split.path and split.path.endswith("/__api__")):
```
nealrichardson, line 37: Since we're always adding `__api__` before we validate, we'll never raise this error in practice

src/posit/connect/urls_test.py
```diff
@@ -1,28 +1,35 @@
...
-def test_fix_without_netloc():
-    with pytest.raises(ValueError):
-        append_path("http:///__api__", "baz")
+def test_append_path_with_leading_slash():
+    assert append_path("http://foo.bar/__api__", "/baz") == "http://foo.bar/__api__/baz"
+
+
+def test_fix_with_correct_url():
+    assert server_to_api_url("http://foo.bar/__api__") == "http://foo.bar/__api__"
+    assert server_to_api_url("http://foo.bar/__api__/") == "http://foo.bar/__api__"
```
nealrichardson, line 16: I noticed this case (with trailing slash) wasn't covered so I added this assertion and then made the test pass.

nealrichardson (approved):
I pushed a couple of changes and just noted them in comments, have a look, otherwise LGTM

PR #46, opened 2024-02-21
test: Add tests for Users with mock responses

nealrichardson:
Plus some TODOs that come out of this (I'll come back to them tomorrow, or you can feel free to push to this branch). This should exercise some of the previously untested code, and should open up a path for adding some enhancements on top.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,4 @@ responses>=0.25
 ruff==0.1.14
 setuptools==69.0.3
 setuptools-scm==8.0.4
+pandas==2.2.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,4 +6,4 @@ responses>=0.25
 ruff==0.1.14
 setuptools==69.0.3
 setuptools-scm==8.0.4
-pandas==2.2.0
+pandas
--- a/src/posit/connect/users_test.py
+++ b/src/posit/connect/users_test.py
@@ -97,7 +97,8 @@ class TestUsers:
 
         con = Client(api_key="12345", url="https://connect.example/")
         # TODO(#48): page_size should go with find(), can't pass it to client.users
-        u = Users(con.config, con.session, page_size=2)
+        u = con.users
+        u.page_size = 2
         # TODO(#47): Add __len__ method to Users
         assert len(u.find().data) == 3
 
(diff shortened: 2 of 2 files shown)


src/posit/connect/users_test.py
```diff
@@ -21,3 +23,79 @@ class TestUsers:
     def test_init(self, mock_config, mock_session):
         with pytest.raises(ValueError):
             Users(mock_config, mock_session, page_size=9999)
+
+    @responses.activate
+    def test_get_users(self):
+        responses.get(
+            "https://connect.example/__api__/v1/users",
```
nealrichardson, line 30: I made these fixtures by querying a real Connect and then sanitizing the responses.

src/posit/connect/users_test.py
```diff
@@ -21,3 +23,79 @@ class TestUsers:
...
+                "current_page": 2,
+                "total": 3,
+            },
+        )
+
+        con = Client(api_key="12345", url="https://connect.example/")
+        # TODO: page_size should go with find(), can't pass it to client.users
+        u = Users(con.config, con.session, page_size=2)
+        # TODO: Add __len__ method to Users
+        assert len(u.find().data) == 3
```
nealrichardson, line 101: What I want this to be is:

```python
u = con.users.find(page_size=2)
assert len(u) == 3
```

tdstein: I was thinking about making `page_size` a `Config` property. There are a handful of endpoints that support paging. Placing the value in Config will make it accessible to all future implementations of `Resources.`

```
       con = Client(api_key="12345", url="https://connect.example/", page_size=2)
       u = con.users.find()
       assert len(u) == 3
```

Adding `page_size` to `find` will require more refactoring.

nealrichardson: > Adding `page_size` to `find` will require more refactoring.

Yeah that's why I didn't just do it yesterday 😂  I had an idea while walking the dog though, may give it a shot later.

nealrichardson: Yeah, my other thought was to ship this, build (or more likely let you build) the `Content` classes, and see how some of this shakes out when we have two kinds of `Resources` to work with.

nealrichardson: I made some issues for these and linked the issue numbers in the comment (i.e. `# TODO(#47)`) so that we can find them later.

requirements-dev.txt
```diff
@@ -6,3 +6,4 @@ responses>=0.25
 ruff==0.1.14
 setuptools==69.0.3
 setuptools-scm==8.0.4
+pandas
```
nealrichardson, line 9: No version pin here bc for our purposes, it doesn't matter, and because the current release of pandas isn't available for Python 3.8.

nealrichardson:
I'm going to merge this: it pushes test coverage to 96% of lines, and it covers the key functionality of the Users resources. I want to try a refactor and need this coverage in place first.

PR #49, opened 2024-02-23
refactor: simplify Users, factor out paginator

### LINKED ISSUE #48: page_size should be argument to find, not __init__
Encountered on #46

nealrichardson:
This PR is a significant refactor, curious what you think. Highlights:

* Reduces the `Users` class to being an interface to the API, not something that holds data or state on its own. This simplifies the result caching story: if you want to work off of a cached set of responses from the API, assign the result of `Users.find()` to something. Otherwise, `client.users` will always query the API fresh. I was starting to get concerned about cache invalidation, as in when you iterate over the collection in order to PUT/PATCH/DELETE things on the server, but your `Users` object would be holding stale state. This also simplifies the Users/CachedUsers split (by eliminating it), which I never could quite wrap my head around. 
* Factors out a `Paginator` class that is designed to work with the Connect API's "[offset pagination](https://docs.posit.co/connect/cookbook/pagination/#offset-pagination)" generally, without being tied to one resource type or another. 
* Deletes the abstract `Resource`, `Resources`, and `CachedResources` classes. I suspect we'll restore something like that in the future, but I think the logic will become clear once we add a second object type (Content, I presume) and see what falls out as common. In the spirit of YAGNI, we don't need it yet.
* Gets test coverage up to 100% of lines.

Behavior changes included here:

* Fixes #48: page_size is an argument to find()
* You can't iterate over `client.users` directly: you have to iterate over `client.users.find()`. I *think* this is probably the right call, explicit better than implicit, etc. But if we later decide otherwise, it's not difficult to add.
* There is no `client.users.to_pandas()` either: you have to `pd.DataFrame(client.users.find())`. Again, this might be the right call, since `find()` returns a `list`, there's not going to be a `.to_pandas()` method on it, so it might be confusing that you can to_pandas on one but not the other.


tdstein:
> Factors out a PaginatedRequester... 

I like the use of `PaginatedRequester` to separate the pagination logic! This is how `boto3` does it, but they make the user do the paging (which I don't like).

> This also simplifies the Users/CachedUsers split (by eliminating it), which I never could quite wrap my head around.

My initial implementation of this isn't great. I got to the CachedResources / Resources split when trying to solve the issue you ran into: "what happens when you do `users.find_one().delete()`...

I was able to reduce the abstraction to a single Resources base class, but I'm still working through the cache invalidation edge cases... This is roughly how I've solved it thus far.

```python 
# On first iteration all users are fetched from the server and cached within Users.
for user in client.users:
    print(user)

# The caller looks for users using server-side filtering arguments. Since server-side filters are supplied, the cached list isn't utilized. But, we can keep it for now.
users: list[users] = client.users.find(/some server side filtering arguments/) 

# The caller does some client-side filtering of the fetched results.
user: User = next(user for user in users if user.foo == 'bar')

# The caller then deletes the user they found.
# This subsequent invalidates the cache in `client.users`. This is accomplished by holding a private reference to the client.users cached list within User.
user.delete()
```

> Gets test coverage up to 100% of lines.

🥳 This is great!

> Fixes https://github.com/posit-dev/posit-sdk-py/issues/48: page_size is an argument to find()

I think it's better *not* to expose this property to the user since max page size (500) is generally more efficient than small pages (maybe YAGNI?). But that's just a hunch.

> You can't iterate over client.users directly...

In the spirit of being idiomatic, `[user for user in client.users]` is it. However, I understand the value of simplifying the implementation to avoid complexity for our sake.

nealrichardson:
> I'm still working through the cache invalidation edge cases

Yeah, I feel like I've been through this before, and the solutions I've come to are (a) cache at the HTTP layer (i.e. your code always does GET, but you get a client-side cached copy if you've done it before, and you invalidate cache when you do PUT/POST/PATCH/DELETE), or (b) don't cache (i.e. make the user be responsible for determining when to go back to the server or not). So my inclination is to start with (b) and add (a) if/when needed. I'm not certain we will need it, depends on the use cases, and honestly I'd rather not worry about pointing guns at our feet unless it becomes an issue.

> In the spirit of being idiomatic, `[user for user in client.users]` is it.

I hear you, and I'm torn on this one, I can see both sides. My instinct says to try the simple/explicit version first (requiring .find()) in practice and see how that feels. Maybe `[user for user in client.users.find()]` will seem natural enough.

mmarchetti:
Could `client.users` implement `iter()` with something like `return self.find().iter()`?

nealrichardson:
> Could `client.users` implement `iter()` with something like `return self.find().iter()`?

Yeah I don't think there's a technical obstacle here, I think it's a question of how much we want the Users object to be directly useful vs. a vehicle to getting to the useful things, and whether it's confusing that Users has some list-like methods on it but has other methods, but `Users.find()` doesn't have those methods because it *is* a list. I can see both sides but was leaning toward the simple side in this PR, figured that's an easier place to start from.

mmarchetti:
Yes. There is plenty of room to add "magic" later.

tdstein:
...why does GitHub let me edit comments. I accidentally overwrote one of yours @nealrichardson...

> I hear you, and I'm torn on this one, I can see both sides. My instinct says to try the simple/explicit version first (requiring .find()) in practice and see how that feels. Maybe `[user for user in client.users.find()]` will seem natural enough. 

I keep getting hung up on `find`. Maybe `client.users.fetch()`, `client.users.all()`, or `client.users.to_list()` (open to other ideas) to delineate the action?

nealrichardson:
Re: exposing page_size or not: I agree that users shouldn't generally want to touch this, and max_page_size should be default, I'm only using it in the tests because I don't want to work with test fixtures with 500+ records. I don't think it's bad to have it as an argument that most folks will never touch.

> ...why does GitHub let me edit comments. I accidentally overwrote one of yours @nealrichardson...

To restate what was lost there: cache invalidation is one of the Two Hard Problems so we should avoid dealing with it unless/until we need to. Whenever that time comes, IME inserting a client-side cache around the HTTP layer is the way to go as that will give us the best chance to invalidate when we need to (but nothing is perfect).


> 
> > I hear you, and I'm torn on this one, I can see both sides. My instinct says to try the simple/explicit version first (requiring .find()) in practice and see how that feels. Maybe `[user for user in client.users.find()]` will seem natural enough.
> 
> I keep getting hung up on `find`. Maybe `client.users.fetch()`, `client.users.all()`, or `client.users.to_list()` (open to other ideas) to delineate the action?

Yeah it's possible that find() isn't the right verb, or that it's good for searching/querying but we want an alias to mean give-me-everything. This is another area that I'm leaning towards shipping as is (i.e. keep it simple) and get some folks trying to use the SDK, see what feedback we get there. I think we've gone back and forth on this enough and haven't seemed to settle on it for ourselves, so maybe we just need input from our target audience.

tdstein:
> Yeah it's possible that find() isn't the right verb, or that it's good for searching/querying but we want an alias to mean give-me-everything. This is another area that I'm leaning towards shipping as is (i.e. keep it simple) and get some folks trying to use the SDK, see what feedback we get there. I think we've gone back and forth on this enough and haven't seemed to settle on it for ourselves, so maybe we just need input from our target audience.

Fair enough. That works for me.

tdstein (approved):

src/posit/connect/pagination.py
```diff
@@ -0,0 +1,49 @@
+from typing import List
+
+
+# The maximum page size supported by the API.
+_MAX_PAGE_SIZE = 500
+
+
+class PaginatedRequester:
```
tdstein, line 8: nit; `Paginator` is a common name for this type of thing.

nealrichardson:
I opened https://github.com/posit-dev/posit-sdk-py/issues/50 to start collecting design questions like these that we're deferring until we can get some feedback from actual use of the SDK.

PR #51, opened 2024-02-23
feat: add client.me property that calls v1/user (lazily)

### LINKED ISSUE #22: SDK client should fetch the current user info at initialization
It would be nice if the SDK populated the client config with some basic info about the current user when it is initialized.

Specifically, for the `OAuthIntegrations` endpoint, I would like the current user's guid to be available with something like `self.config.current_user.guid`, assuming my endpoint looks something like this:

```python
class OAuthIntegration:
    def __init__(
        self, config: Config, session: Session, *args
    ) -> None:
        super().__init__()
        self.config = config
        self.session = session

    def get_credentials(self, content_identity: Optional[str]) -> Response:
        endpoint = os.path.join(self.config.endpoint, "v1/oauth/integrations/credentials")

        # craft a basic credential exchange request where the self.config.api_key owner
        # is requesting their own credentials
        data = dict()
        data["grant_type"] = "urn:ietf:params:oauth:grant-type:token-exchange"
        data["subject_token_type"] = "urn:posit:connect:user-guid"
        data["subject_token"] = self.config.current_user.guid

        # if this content is running on Connect, then it is allowed to request
        # the content viewer's credentials
        if content_identity:
            data["subject_token"] = parse_viewer_guid(content_identity)
            data["actor_token_type"] = "urn:posit:connect:content-identity-token"
            data["actor_token"] = content_identity

        return self.session.post(endpoint, data=data)
```

nealrichardson:
Replacement for #33, was easier to pull the changes to a fresh branch than try to rebase

Fixes #22.

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -7,7 +7,7 @@ from . import hooks, urls
 
 from .auth import Auth
 from .config import Config
-from .users import Users
+from .users import User, Users
 
 
 class Client:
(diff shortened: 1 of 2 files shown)


src/posit/connect/client.py
```diff
@@ -35,14 +35,23 @@ def __init__(
         # Store the Session object.
         self.session = session
 
-        # Place to cache the server settings
-        self.server_settings = None
+        # Internal attributes to hold settings we fetch lazily
+        self._server_settings = None
+        self._current_user: Optional[User] = None
```
tdstein, line 40: Since we are moving away from caching in `Users`, should we remove it here for consistency?

tdstein (approved):

src/posit/connect/client.py
tdstein: Saving the GUID seems safe, but other information regarding the `current_user` could change between calls to `client.me`.

This morning, at least, I like the idea of not caching anything by convention. We can add it back in if the server request load becomes burdensome.

nealrichardson: Yep, I agree. I'll remove, then merge.

PR #52, opened 2024-02-23
Adds an OAuthIntegration 

dbkegley:
Adds a `client.oauth` resource for interacting the the `v1/oauth/integrations/credentials` endpoint to perform a token exchange with Connect.


src/posit/connect/client.py
```diff
@@ -37,13 +46,20 @@ def __init__(
...
     def connect_version(self):
         if self.server_settings is None:
             self.server_settings = self.get("server_settings").json()
         return self.server_settings["version"]
 
+    @property
+    def oauth(self) -> OAuthIntegration:
+        if self._oauth is None:
+            self._oauth = OAuthIntegration(config=self.config, session=self.session)
+        return self._oauth
```
nealrichardson, line 61: AFAICT there's nothing in `OAuthIntegration` that needs to be cached, so we just need to be able to access it here, as we do (now) for `Users` below.

```suggestion
        return OAuthIntegration(config=self.config, session=self.session)
```

src/posit/connect/oauth.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(
+        self, config: Config, session: Session
+    ) -> None:
+        self.url = urls.append_path(config.url, "v1/oauth/integrations/credentials")
+        self.config = config
+        self.session = session
+
+
+    def get_credentials(self, user_identity: Optional[str]) -> Response:
```
nealrichardson, line 20: Does this need to return `Response` or can it return the response body or something? I don't see this being used anywhere yet, so I can't tell.

I think we're trying to avoid having the request/response details leaking outside of the get/put/post/etc. methods.

src/posit/connect/client.py
```diff
@@ -1,15 +1,24 @@
...
 
 from . import hooks, urls
 
 from .auth import Auth
 from .config import Config
+from .oauth import OAuthIntegration
 from .users import Users
 
 
+# Connect sets the value of the environment variable RSTUDIO_PRODUCT = CONNECT
```
nealrichardson, line 15: 😭  it's 2024, we should start setting `POSIT_PRODUCT` in the env too, yeah?

nealrichardson (commented):
A few notes. LMK if you want to talk through using `responses` to set up some tests that stub out the server side.

src/posit/connect/oauth.py
nealrichardson: Ah sorry, I was using github's symbol lookup and it didn't find it, should have stuck with good ol' cmd-F. 

Yeah in that case, I'd have this method return `response.json()`

tdstein: A defined type would be even better so that the consumer gets type hints.

```python
@dataclass
class OAuthResponse:
    access_token: str
```

src/posit/connect/external/databricks.py
```diff
@@ -0,0 +1,60 @@
...
+# https://github.com/databricks/databricks-sdk-py/blob/v0.20.0/databricks/sdk/credentials_provider.py
+# https://github.com/databricks/databricks-sql-python/blob/v3.1.0/src/databricks/sql/auth/authenticators.py
+# In order to keep compatibility with the Databricks SDK
+class CredentialsProvider(abc.ABC):
+    """CredentialsProvider is the protocol (call-side interface)
+     for authenticating requests to Databricks REST APIs"""
+
+    @abc.abstractmethod
+    def auth_type(self) -> str:
+        ...
```
tdstein, line 18: Should these abstract method's `raise NotImplementedError()`?

src/posit/connect/client.py
```diff
@@ -1,15 +1,24 @@
...
+# when content is running on a Connect server. Use this var to determine if the
+# client SDK was initialized from a piece of content running on a Connect server.
+def is_local() -> bool:
+    return not os.getenv("RSTUDIO_PRODUCT") == "CONNECT"
```
tdstein, line 19: It looks like this is only being used [src/posit/connect/external/databricks.py](https://github.com/posit-dev/posit-sdk-py/pull/52/files#diff-b826271d8297d4f972ef39df49328e78f4a7fada12ebe598b587cf42160eaae8). 

Should it be moved there?

This could also go in `Config` if there is reuse. But I'd rename it to something like `run_mode`, `runtime_environment`, etc.

src/posit/connect/external/databricks.py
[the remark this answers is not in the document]
dbkegley: Here's our example from last weeks demo. 

https://github.com/posit-dev/posit-sdk-py/blob/62fcc8e81574adcc69be1a4a03a0e5ab4831687d/connect-content/sample-content.py#L27-L30

This usage has changed a little bit since then. In this PR we accept a `client` as an optional parameter so that the Content doesn't require a Connect server when running locally. The SDK client is initialized only when the content is running on Connect by default.

src/posit/connect/external/databricks.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ..client import Client, is_local
+from ..oauth import OAuthIntegration
+
+HeaderFactory = Callable[[], Dict[str, str]]
+
+# https://github.com/databricks/databricks-sdk-py/blob/v0.20.0/databricks/sdk/credentials_provider.py
+# https://github.com/databricks/databricks-sql-python/blob/v3.1.0/src/databricks/sql/auth/authenticators.py
+# In order to keep compatibility with the Databricks SDK
+class CredentialsProvider(abc.ABC):
```
tdstein, line 12: Why not import https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/credentials_provider.py#L30 directly?

src/posit/connect/external/databricks.py
```diff
@@ -0,0 +1,60 @@
...
+        return None
+
+    if client is None:
+        client = Client()
+
+
+    # If the user-identity-token wasn't provided and we're running on Connect then we raise an exception.
+    # user_identity is required to impersonate the viewer.
+    if user_identity is None:
+        raise Exception("The user-identity-token is required for viewer authentication.")
```
tdstein, line 54: `raise ValueError` here instead of `Exception`

src/posit/connect/external/databricks.py
```diff
@@ -0,0 +1,60 @@
...
+        return "posit-oauth-integration"
+
+    def __call__(self, *args, **kwargs) -> HeaderFactory:
+        def inner() -> Dict[str, str]:
+            access_token = self.posit_oauth.get_credentials(self.user_identity).json()['access_token']
+            return {"Authorization": f"Bearer {access_token}"}
+        return inner
+
+
+def viewer_credentials_provider(client: Optional[Client], user_identity: Optional[str]) -> Optional[CredentialsProvider]:
```
tdstein, line 40: Is the value of user_identity ever something other than `headers.get('Posit-Connect-Content-Identity')`?

src/posit/connect/client.py
tdstein: 👍🏻 Place it where it's used for now.

src/posit/connect/external/databricks.py
dbkegley: I wanted to avoid adding their SDK as a dependency. I'll test and see if we can remove this abstract definition altogether.

src/posit/connect/oauth.py
atheriel: With the expectation that this endpoint might eventually return something other than an `access_token`, I like the idea of returning a defined type.

src/posit/connect/client.py
atheriel: Can we use the presence of `CONNECT_SERVER` instead and sidestep branding issues?

src/posit/connect/external/databricks.py
dbkegley: Let's leave this for now. I think we will have a similar problem for our content types. It would be nice to provide helpers for each framework when obtaining the identity token from the headers but it will vary by content type.

src/posit/connect/external/databricks.py
```diff
@@ -0,0 +1,66 @@
...
+        def inner() -> Dict[str, str]:
+            access_token = self.posit_oauth.get_credentials(self.user_identity)['access_token']
+            return {"Authorization": f"Bearer {access_token}"}
+        return inner
+
+
+# Use these environment varariables to determine if the
+# client SDK was initialized from a piece of content running on a Connect server.
+def is_local() -> bool:
+    return not os.getenv("CONNECT_SERVER") and not os.getenv("CONNECT_CONTENT_GUID")
```
nealrichardson, line 44: I don't think `CONNECT_SERVER` is right here: that's what we recommend folks set locally in all of our API examples (e.g. https://docs.posit.co/connect/cookbook/). 

I'd rather use the accurate variable even if it says `RSTUDIO`; we can fix that on the server going forward and (eventually) purge it here.

PR #53, opened 2024-02-24
feat: initial support for getting content

nealrichardson:
I was curious to see how smooth it would be to add a new collection, following the pattern of Users and using `responses` for mocking. It was pretty smooth--this took me maybe 45 minutes, about half of that just contemplating the API and how it's different from v1/users, how it has additional query params we'll want to expose, etc. 

For the record, my workflow was:

* Create empty files: content.py and content_test.py
* Use `client.get("v1/content")` to retrieve responses from a real Connect, try to cherry-pick a few items from that that had some variation to use as test fixtures
* Sanitize those: manipulate GUIDs and other ids, replace owner_guids with ids of the users from users_test, alter names
* Set up the test around that, using `responses.get` with that, then add test of `client.content.find()`, which fails of course because client has no attribute content yet
* Add the property to Client, then copy the contents of users.py to content.py and adapt it. This was mostly replacing class names and stripping out pagination, since `v1/content` is not paginated 🤯 
* Fill in the rest of the tests


nealrichardson:
I'm going to merge this so that I can help stub out something for Brian's new search endpoint. Happy to revisit any of the decisions here whenever you get a chance to review, or in future PRs.

PR #59, opened 2024-02-29
feat: adds iterable support to pagination

tdstein:
I'm working on a change to make Paginator a mixin which adds iterable support. Currently, the paginator is a property of Users, but the end state will be `class Users(Paginator[User])`. This will add support to Users for `iter`, `next`, and `len` builtins.

### DIFF
--- a/src/posit/connect/paginator.py
+++ b/src/posit/connect/paginator.py
@@ -16,7 +16,7 @@ class Paginator(Iterable):
     Attributes:
         elements (List[dict]): The list of fetched elements.
         exhausted (bool): Indicates whether all pages have been fetched.
-        idx (int): The current index of the elements list.
+        index (int): The current index of the elements list.
         page_size (int): The number of items to fetch per page.
         session (requests.Session): The requests session object to use for making API requests.
         url (str): The URL of the API endpoint to fetch paginated results from.
@@ -25,7 +25,7 @@ class Paginator(Iterable):
     def __init__(self, session: requests.Session, url: str, *, page_size: int) -> None:
         self.elements: List[dict] = []
         self.exhausted = False
-        self.idx = 0
+        self.index = 0
         self.page_size = page_size
         self.session = session
         self.url = url
@@ -37,7 +37,7 @@ class Paginator(Iterable):
         Returns:
             Iterator[dict]: An iterator object that iterates over the items in the Paginator.
         """
-        self.idx = 0
+        self.index = 0
         return self
 
     def __len__(self) -> int:
@@ -66,19 +66,19 @@ class Paginator(Iterable):
         Returns:
             dict: The next element in the paginator.
         """
-        if self.idx >= len(self.elements):
+        if self.index >= len(self.elements):
             if self.exhausted:
                 raise StopIteration
 
-            page = int(self.idx / self.page_size) + 1
+            page = int(self.index / self.page_size) + 1
             results, self.exhausted = self.fetch(page)
             if not results:
                 raise StopIteration
 
             self.elements += results
 
-        v = self.elements[self.idx]
-        self.idx += 1
+        v = self.elements[self.index]
+        self.index += 1
         return v
 
     def fetch(self, page: int) -> tuple[Iterator[dict] | None, bool]:
--- a/src/posit/connect/paginator.py
+++ b/src/posit/connect/paginator.py
@@ -1,5 +1,5 @@
 from collections.abc import Iterable
-from typing import Iterator, List
+from typing import Iterator, List, Tuple
 
 import requests
 
@@ -81,7 +81,7 @@ class Paginator(Iterable):
         self.index += 1
         return v
 
-    def fetch(self, page: int) -> tuple[Iterator[dict] | None, bool]:
+    def fetch(self, page: int) -> Tuple[Iterator[dict], bool]:
         """
         Fetches a specific page of results from the API.
 
--- a/src/posit/connect/paginator.py
+++ b/src/posit/connect/paginator.py
@@ -44,18 +44,12 @@ class Paginator(Iterable):
         Raises:
             ValueError: If the page number is not greater than 0.
         """
-        # Check if page is greater than 0
         if not (page > 0):
             raise ValueError(f"page must be greater than 0, got {page}")
-        # Define query parameters for pagination.
         params = {"page_number": page, "page_size": self.page_size}
-        # Send a GET request to the endpoint with the specified parameters.
         response = self.session.get(self.url, params=params)
-        # Convert response to dict
         as_dict: dict = dict(response.json())
-        # Parse the JSON response and extract the results.
         results = as_dict["results"]
-        # Mark exhausted if the result size is less than the maximum page size.
         exhausted = len(results) < self.page_size
         return (results, exhausted)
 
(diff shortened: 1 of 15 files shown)


tdstein:
Once this is stable, it might be a good time to add the rest of the paginated API endpoints.

tdstein:
It looks like there are 6 total. Searching for page_number here: https://docs.posit.co/connect/api/swagger.json

src/posit/connect/paginator.py
```diff
@@ -1,49 +1,111 @@
-from typing import List
+from collections.abc import Iterable
+from typing import Iterator, List, Tuple
 
+import requests
 
-# The maximum page size supported by the API.
-_MAX_PAGE_SIZE = 500
```
nealrichardson, line 5: Is this not constant across all paginated APIs?

src/posit/connect/paginator.py
```diff
@@ -1,49 +1,111 @@
...
+    def __len__(self) -> int:
+        """
+        Returns the total number of items in the paginator.
+
+        Returns:
+            int: The total number of items in the paginator.
+        """
+        # Define query parameters for pagination.
+        params = {"page_number": 1, "page_size": 1}
+        # Send a GET request to the endpoint with the specified parameters.
```
nealrichardson, line 81: I think you can delete these prompt-comments, they don't add any useful to the code

src/posit/connect/paginator.py
```diff
@@ -1,49 +1,111 @@
...
+        return self
+
+    def __len__(self) -> int:
+        """
+        Returns the total number of items in the paginator.
+
+        Returns:
+            int: The total number of items in the paginator.
+        """
+        # Define query parameters for pagination.
```
nealrichardson, line 79:
```suggestion
        # Connect's paginated APIs include a "total" count in the response.
        # So we only need to request page_size = 1 and inspect the total.
```

src/posit/connect/paginator.py
```diff
@@ -1,49 +1,111 @@
...
+        params = {"page_number": 1, "page_size": 1}
```
nealrichardson, line 80: Does `page_size=0` work?

src/posit/connect/users.py
```diff
@@ -25,32 +25,73 @@ class User(TypedDict, total=False):
...
+        *,
+        page_size: int = _MAX_PAGE_SIZE,
+    ) -> Optional[User]:
+        # You may be tempted to to use self.find(filter, page_size) and return the first element.
+        # This is less efficient as it will fetch all the users and then filter them.
+        # Instead, we use the paginator directly to fetch the first user that matches the filter.
+        # Since the paginator fetches the users in pages, it will stop fetching subsequent pages once a match is found.
+        url = urls.append_path(self.config.url, "v1/users")
+        paginator = Paginator(self.session, url, page_size=page_size)
+        users = (User(**user) for user in paginator)
```
nealrichardson, line 90: Reading this, it's not obvious to me that each of these comprehensions isn't actually iterating over all of the results in the paginator. (There at least used to be versions of Python where that wouldn't be true, unless I'm misremembering.)

So for me, that merits (a) an inline comment acknowledging how it works, and (b) a `responses`-based test that would error if you went beyond a certain page of results. The current test doesn't ensure that, IIRC.

nealrichardson (commented):
A few notes here.

This is cool, but I'm wondering how it helps. To my eyes, it's less immediately obvious how the pagination works in practice with this implementation--the previous one is much less clever. So what do we get for this extra complexity, and is it worth it? It may very well be worth it, I just thought it worth asking.

src/posit/connect/paginator.py
tdstein: Yeah, it's interesting. I've done this a few times in different places and have seen mixed results. Some have found them very useful, and others haven't. You can get the same information from `/explain` in GitHub Copilot, but that's a new workflow concept for most folks.

src/posit/connect/paginator.py
tdstein: Unfortunately, no, `posit.connect.errors.ClientError: The parameter page_size=0 is less than the allowed minimum 1 (Error Code: 25, HTTP Status: 400 Bad Request)`.

src/posit/connect/users.py
tdstein: I added some comments to try and explain the behavior.

tdstein:
> A few notes here.
> 
> This is cool, but I'm wondering how it helps. To my eyes, it's less immediately obvious how the pagination works in practice with this implementation--the previous one is much less clever. So what do we get for this extra complexity, and is it worth it? It may very well be worth it, I just thought it worth asking.

Thanks! I want to avoid unnecessary complexity, so happy to iron this out more.

Could you expand on which parts you find complex?

I have a few goals in mind.

1. Allow idiomatic comprehension on resource collections: `(user for user in client.users)`.
1. Provide mixin support for other resource types.
1. Make the comprehension memory efficient by waiting to fetch subsequent pages until they are needed.

As we've been discussing. The memory efficiency isn't premature since endpoint support is limited. But I'm hoping that this can encourage the migration of other endpoints which manage large collections.

src/posit/connect/paginator.py
```diff
@@ -1,49 +1,103 @@
...
+        return self
+
+    def __len__(self) -> int:
+        """
+        Returns the total number of items in the paginator.
+
+        Returns:
+            int: The total number of items in the paginator.
+        """
+        # Connect's paginated APIs include a "total" count in the response.
```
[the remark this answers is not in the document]
tdstein: Thanks for pointing this out. I wasn't aware of this API yet. Yes, we can create two different `Paginatior` implementations. Add https://github.com/posit-dev/posit-sdk-py/issues/75

tdstein:
> * I don't think that it's safe to lazily take pages to feed the iterator. The content API at least is sorted by "recently updated", right? So if I iterate through that, make updates to some or all of the items in it, and fetch pages lazily, the contents of e.g. "page_number=2" may be changing as a result of my action. Users seem to be sorted alphabetically. So if paginated results are sorted on mutable properties, we may get surprising results unless we fetch all before allowing actions that could mutate the server. Even if sorting on immutable properties, the ability to delete entries will make page contents unstable.

I see this as a server-side implementation issue rather than a client-side one. Without consistent ordering and support for pagination tokens, the underlying dataset can change regardless of the time between page fetches.

nealrichardson:
> > * I don't think that it's safe to lazily take pages to feed the iterator. The content API at least is sorted by "recently updated", right? So if I iterate through that, make updates to some or all of the items in it, and fetch pages lazily, the contents of e.g. "page_number=2" may be changing as a result of my action. Users seem to be sorted alphabetically. So if paginated results are sorted on mutable properties, we may get surprising results unless we fetch all before allowing actions that could mutate the server. Even if sorting on immutable properties, the ability to delete entries will make page contents unstable.
> 
> I see this as a server-side implementation issue rather than a client-side one. Without consistent ordering and support for pagination tokens, the underlying dataset can change regardless of the time between page fetches.

Sure, it's a server-side issue, but supposing we change/fix it there, the SDK still has to support the current API. We can only hope to minimize its impact by fetching all at the ~same time (AFAICT).

tdstein:
I'm going to abandon this idea for the time being. We can revisit if we decide to make resources iterable in the future.

PR #66, opened 2024-03-04
build: adds fix to pre-commit hooks

tdstein:


Makefile
```diff
@@ -45,6 +46,9 @@ deps:
 fmt:
 	$(PYTHON) -m ruff format .
 
+fix:
```
nealrichardson, line 49: I think you added an enumeration of all of the Makefile commands to the CONTRIBUTING.md last week, should this be added there? (Alternatively, you could delete the list from CONTRIBUTING, seems like it is destined to get out of sync, and this Makefile isn't too hard to scan to see what it does.)

tdstein: I'll edit the CONTRIBUTING to direct readers to look at the Makefile.

PR #70, opened 2024-03-04
ci: adds action to lint conventional commits.

tdstein:
@nealrichardson and I are using conventional commits to keep the Git changelog clean. This change adds a GitHub action to verify that the pull request title uses the conventional commit syntax in the title. 

Since we are forcing "Squash and merge", developers can continue using whatever commit title they want for individual commits. By default, those commits will be squashed into the commit body.

To test the GitHub action functionality, try fixing the title of this pull request to match the style guide.

### DIFF
--- a/.github/workflows/conventional-commits.yaml
+++ b/.github/workflows/conventional-commits.yaml
@@ -1,0 +1,50 @@
+name: Conventional Commits
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+jobs:
+  default:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        id: lint
+        env:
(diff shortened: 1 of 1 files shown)


nealrichardson:
> I like this pattern a lot - it'll both make it easier to grok the commit history and opens up the possibility of scripting interactions with our git history. Very cool.
> 
> Do we need something in place to guide how commit types can (if they should) shift over time? Say we find certain commit types are noisy, or we need one that isn't enumerated here.

IIUC this only runs on new PRs, so if we want to modify the list of commit types going forward, we just edit the workflow.

nealrichardson (approved):
I like this idea! No opinion about the specifics of the GHA workflow, but we can always revise if needed.

tdstein:
> I like this pattern a lot - it'll both make it easier to grok the commit history and opens up the possibility of scripting interactions with our git history. Very cool.
> 
> Do we need something in place to guide how commit types can (if they should) shift over time? Say we find certain commit types are noisy, or we need one that isn't enumerated here.

Thanks for taking a look! The commit types are configurable via the GitHub Action. As @nealrichardson mentioned, we can modify these as scenarios arise.

Issue #54, opened 2024-02-27
get_current_user


edavidaja:
Shiny supports this via [`session.user`](https://docs.posit.co/connect/user/shiny-python/#user-meta-data), for other frameworks we should support this via SDK:

- https://docs.posit.co/connect/user/fastapi/#user-meta-data
- https://docs.posit.co/connect/user/flask/#user-meta-data
- https://docs.posit.co/connect/user/dash/#user-meta-data
- https://docs.posit.co/connect/user/streamlit/#user-meta-data

nealrichardson:
We just added https://github.com/posit-dev/posit-sdk-py/pull/51. So should we plan to update those guides to use the SDK to get the user? Or something else?

mmarchetti:
The API call in #51 will return the user who owns the At the client is configured to use. The scenario here is a deployed app that wants to know the identity of visiting users who are identified to Connect (but not the app) via a session cookie. Connect embeds their information in an HTTP header that the app receives. Code to extract HTTP headers from the inbound request is framework-specific.

tdstein:
Is `visitor` the appropriate name for this entity? While `user` for "me".

e.g.,
```python
client.visitor
```

edavidaja:
> So the SDK would be providing a helper function to encapsulate that?

That was what I had in mind.

tdstein:
Got it. Is `visitor` still appropriate for a logical name?

I'm thinking the following where get_visitor is the tbd helper method

```python
import json

from posit.connect import Client

def get_visitor() -> dict:
    return {'username': 'taylor_steinberg', 'groups': ['rsc_team']}

with Client() as client:
    visitor = get_visitor()
    user = client.users.find_one(lambda user: user["username"] == visitor["username"])
    print(json.dumps(user))
```

edavidaja:
I think "visitor" is fine and it's a word we use in other places in the documentation. A good alternative is "viewer"--it overlaps with the Connect role, which I think is fine in this case.

mmarchetti:
> ```python
> visitor = get_visitor()
> ```

Looking at our example code for [Flask](https://docs.posit.co/connect/user/flask/#user-meta-data), [FastAPI](https://docs.posit.co/connect/user/fastapi/#user-meta-data) and [Streamlit](https://docs.posit.co/connect/user/streamlit/#user-meta-data), it will be difficult for the API to have that shape since we don't know which framework is in use. We might need a separate helper function per framework.

PR #76, opened 2024-03-05
docs: Added Dash example.

plascaray:


examples/connect/dash/app.py
```diff
@@ -0,0 +1,78 @@
...
+)
+
+def update_page(_):
+    """
+    Dash example application that shows user information and 
+    the first few rows from a table hosted in Databricks.
+    """
+
+    session_token = flask.request.headers.get('Posit-Connect-User-Session-Token')
+    credentials_provider = viewer_credentials_provider(user_session_token=session_token)
```
nealrichardson, line 43: I see a pattern like this in all of the examples, and I wonder if we can wrap this up more. Can we have a single function that detects, or try/excepts, to get the header from the places where shiny/flask/etc. put it? (Tangentially related: #54)

dbkegley: I've had the same thought. I think that would be a nice addition. We'll need to follow the same pattern we did with `pandas` to avoid having our SDK depend on each of the frameworks.

```python
try:
  import flask
except ImportError:
  return None
```

It also varies quite a bit depending on the framework. For flask/fastapi, the headers are on the incoming request object. For shiny it's on the `session` which is only available in the server block. We'll have to experiment with each of the frameworks more to see what feels like the right interface for this helper.

dbkegley (approved):

examples/connect/dash/README.md
```diff
@@ -0,0 +1,37 @@
...
+```bash
+rsconnect version
+```
+
+Or install it as documented in the [installation](https://docs.posit.co/rsconnect-python/#installation) section of the documentation.
+
+To publish, make sure `CONNECT_SERVER`, `CONNECT_API_KEY`, `DATABRICKS_HOST`, `DATABRICKS_PATH` have valid values. Then, on a terminal session, enter the following command:
+
+```bash
+rsconnect deploy shiny . \
```
jonkeane, line 22:
```suggestion
rsconnect deploy dash . \
```

Should this be `dash` here?

examples/connect/dash/README.md
```diff
@@ -0,0 +1,37 @@
...
+  --environment DATABRICKS_HOST \
+  --environment DATABRICKS_PATH
+```
+
+Note that the Databricks environment variables do not need to be resolved by the shell, so they do not include the `$` prefix.
+
+The Databricks environment variables only need to be set once, unless a change needs to be made. If the values have not changed, you don’t need to provide them again when you publish updates to the document.
+
+```
+rsconnect deploy shiny . \
```
jonkeane, line 34:
```suggestion
rsconnect deploy dash . \
```

Same as above?

examples/connect/dash/app.py
nealrichardson: > For shiny it's on the session which is only available in the server block

Slighly complicated further by the fact that Shiny Express doesn't have a server block. But still, may be worth exploring.

examples/connect/dash/README.md
```diff
@@ -0,0 +1,37 @@
+# Dash Example
+
+## Start the app locally
+
+```bash
+python app.py
```
dbkegley, line 6: I think we should talk about the required environment variables before we show how to run the app. This fails if the user hasn't set `DATABRICKS_HOST` and `DATABRICKS_PATH` in their shell

PR #78, opened 2024-03-06
docs: update streamlit example and readmes

dbkegley:
This PR makes the streamlit example consistent with our other examples

I'm also going to update the other PRs to talk about the required environment vars before running the apps locally


examples/connect/streamlit/requirements.txt
```diff
@@ -1,56 +1,4 @@
-altair==5.2.0
-attrs==23.2.0
-blinker==1.7.0
-cachetools==5.3.2
-certifi==2024.2.2
-charset-normalizer==3.3.2
-click==8.1.7
+databricks-sql-connector==3.0.1
```
jonkeane, line 1: This is more a question than an issue or suggestion: what happened with all of these dependencies? Were they old copy 🍝  or something?

examples/connect/streamlit/app.py
```diff
@@ -12,36 +12,27 @@
...
-if headers:
-    USER_SESSION_TOKEN = headers.get("Posit-Connect-User-Session-Token")
-
-credentials_provider = viewer_credentials_provider(
-    user_session_token=USER_SESSION_TOKEN
+cfg = Config(
+    host=DATABRICKS_HOST_URL, credentials_provider=credentials_provider
 )
-cfg = Config(host=DB_HOST_URL, credentials_provider=credentials_provider)
-# cfg = Config(host=DB_HOST_URL, token=DB_PAT)
```
jonkeane, line 32: ❤️  🧹 #️⃣ 

Thank for cleaning this up

examples/connect/streamlit/requirements.txt
dbkegley: I did a pip freeze initially on this example but I like the approach that @plascaray used in his examples better where we only capture the dependencies that we really need, and then let pip handle transitive deps. No need for us to record them here IMO so I was cleaning this up so that it's consistent with the other examples

jonkeane: 💯 Absolutely 

Thanks!

Issue #23, opened 2024-02-15
Expected usage for Resource objects that need custom methods?


dbkegley:
I see in the readme that we can operate on Resources using the methods defined by the base Resource classes, like `find()`, `find_one()`, `get()`, `delete()` etc.  What should the behavior be for resources where we want to add custom methods that don't fit into one of the standard CRUD resource operations. 

For example:

```python
from .resources import Resource

class OAuthIntegration(Resource):
    name: str

    def get_credentials(self): # throws: TypedDict classes can contain only type annotations
        pass
```

edit:

I could see this also being relevant for an Rmd that I wanted to re-render. e.g.

```python
con.content.get("1234-5678-90ab-cdef").render()
```

nealrichardson:
Poking at this some more, `dataclass` seems like it could work, though I think we'd have to think through the backwards/forwards compatibility implications (e.g. if the entities grow extra attributes in the API responses, or if you're querying an older Connect that doesn't have all fields). Other options that occur to me that may be more flexible, for better and worse:

* Inherit from `dict` and add `@property`s for accessing the fields (with type annotations). 
* Don't inherit from dict, put the raw API response in `self._data`, and use `@property`s to access them.

tdstein:
> if the entities grow extra attributes in the API responses, or if you're querying an older Connect that doesn't have all fields

This is the caveat that led me down the TypedDict path. 

I'm satisfied as long as we can support conversion from `dict` (.e.g., `response.json()`) to `Resource`.

nealrichardson:
> I'm satisfied as long as we can support conversion from `dict` (.e.g., `response.json()`) to `Resource`.

I think that's the easy part. So let's summarize the alternatives:

:x: subclass `TypedDict`: can't add methods
:x: use `dataclass`: too rigid to accommodate API changes
:x: subclass `dict`: I seemed to remember this being a bad idea usually, and [this](https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/) pointed out that we'll have trouble when defining a new method for `update()` since that's a base dict method too.
🔲  subclass `UserDict`: mentioned in that blog post, should investigate, sounds most promising of remaining alternatives
🔲  subclass `MutableMapping`: probably overkill to reimplement all of those methods
🔲  regular `object`, put data in `_data`: I'd like to preserve `pd.DataFrame([list of objects])` and I'm not sure that would work, could investigate or pursue alternatives.

[...]

PR #83, opened 2024-03-06
feat: adds base class definition for resources

tdstein:
Adding an ABC class to enforce the style choices defined in ./src/connect/README.md

### DIFF
--- a/src/posit/connect/resources.py
+++ b/src/posit/connect/resources.py
@@ -1,0 +1,31 @@
+from abc import ABC, abstractmethod
+from typing import Generic, List, Optional, TypeVar
+
+
+T = TypeVar("T")
+
+
+class Resources(ABC, Generic[T]):
+    @abstractmethod
+    def create(self, *args, **kwargs) -> T:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def delete(self, *args, **kwargs) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def find(self, *args, **kwargs) -> List[T]:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def find_one(self, *args, **kwargs) -> Optional[T]:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get(self, *args, **kwargs) -> T:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def update(self, *args, **kwargs) -> T:
+        raise NotImplementedError()
--- a/src/posit/connect/users.py
+++ b/src/posit/connect/users.py
@@ -5,10 +5,12 @@ from typing import Callable, List, TypedDict
 
 from requests import Session
 
+
 from . import urls
 
 from .config import Config
 from .paginator import _MAX_PAGE_SIZE, Paginator
+from .resources import Resources
 
 
 class User(TypedDict, total=False):
@@ -25,7 +27,7 @@ class User(TypedDict, total=False):
     locked: bool
 
 
-class Users:
+class Users(Resources[User]):
     def __init__(self, config: Config, session: Session) -> None:
         self.url = urls.append_path(config.url, "v1/users")
         self.config = config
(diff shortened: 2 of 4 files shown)


src/posit/connect/resources.py
```diff
@@ -0,0 +1,31 @@
...
+T = TypeVar("T")
+
+
+class Resources(ABC, Generic[T]):
+    @abstractmethod
+    def create(self, *args, **kwargs) -> T:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def delete(self, *args, **kwargs) -> None:
```
nealrichardson, line 14: Does `delete()` make sense for these collections? Would this be a bulk delete action?

src/posit/connect/resources.py
```diff
@@ -0,0 +1,31 @@
...
+    @abstractmethod
+    def find(self, *args, **kwargs) -> List[T]:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def find_one(self, *args, **kwargs) -> Optional[T]:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get(self, *args, **kwargs) -> T:
```
nealrichardson, line 26: Should `id`/`guid` be a part of the `get()` signature?

src/posit/connect/users.py
```diff
@@ -25,7 +27,7 @@ class User(TypedDict, total=False):
     locked: bool
 
 
-class Users:
+class Users(Resources[User]):
```
nealrichardson, line 30: What use is it to have `Resources[User]` vs. a simple `Resources` base class? Can/should you use the `T` type in the method definitions? Do you need to define the return types for the methods, or are they redundant to the base class?

More meta question: supposing you could factor some of those things out of the method definitions, should you? I could see an advantage of the current version being that it is explicit (famously better than implicit in Python ;), that you can read the class definition and see what it's doing.

src/posit/connect/resources.py
```diff
@@ -0,0 +1,31 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def update(self, *args, **kwargs) -> T:
```
nealrichardson, line 30: Same question as delete: would this be for bulk updates (regardless of whether the APIs would force us to do one request per entity)?

src/posit/connect/resources.py
tdstein: Maybe? My assumption here is that certain may want to use other argument parameters.

src/posit/connect/users.py
tdstein: I'm not sure I follow. There may be something I need to learn about generics in Python.

Using a parametrized type in the ABC enforces that the inheriting class has consistent return types across methods. Are you suggesting that there are scenarios where this won't be the case?

nealrichardson: I guess one place I'm confused is that we're still defining the return types in this subclass (e.g. L38, find() returns `List[User]`). Can/should those type annotations be removed because the base class will handle them for you?

My other question is: for example, on L40, we call `User(**user)`, can/should that be `T(**user)`?

The other way I've seen this done is that you would set `self._resource_type = User` or something and use that to construct the objects. (Not saying that's worth doing necessarily.)

tdstein: In this case, consider `Resources` an interface instead of an abstract class since it only has abstract methods. If `Resources` contained non-abstract methods, then `T(**user)` should be the correct usage for conversion in those methods (although it would be `T(**result)`). 

I believe `self._resource_type = User` pattern was correct before generics were introduced in Python. This would have been the only way to inform the base class of the implementing type.

> Can/should those type annotations be removed because the base class will handle them for you?

They can, but when writing the implementing code, the Python language server will auto-complete with the correct type definitions.

src/posit/connect/resources.py
tdstein: Based on this method definition, the implementation can do either. We haven't tried injecting the session object into the associated Resource yet. This should be solved as part of #23.

Typically, I would expect this method to act on a collection, which would either be the syntax sugar of `user.delete() for user in user` (pseudo-code) or a direct call to a DELETE endpoint that supports deleting multiple items in a collection.

nealrichardson (approved):

PR #85, opened 2024-03-08
style: Remove unneccesary method in FastAPI example

hekhuisk:
I was looking at this example and noticed the `get_token` method wasn't really necessary since it was just returning what it was given. If you want to keep the method though it can be simplified to just `return posit_connect_user_session_token`. The if statement is not needed.

### DIFF
--- a/examples/connect/fastapi/app.py
+++ b/examples/connect/fastapi/app.py
@@ -8,7 +8,7 @@ from databricks import sql
 
 from typing import Annotated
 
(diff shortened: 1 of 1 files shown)


nealrichardson:
> Thanks @hekhuisk!
> 
> You can ignore the error in Python 3.12 testing. There is a configuration error when running on forked PRs.

I made https://github.com/posit-dev/posit-sdk-py/issues/87 for fixing this. 

Thanks @hekhuisk!

PR #86, opened 2024-03-08
chore: Factor some mock responses to JSON files and read them in with a helper

nealrichardson:
The goals here are to make the test files more readable by separating out the JSON API responses to separate files, and also to make it more natural to reuse the responses in multiple tests. I've just done this for the users mocks right now so you can review what it looks like before I go further.

It may be easier to review the commits individually. The first commit is the refactor and setup for that. In the second commit, I used it to separate out the users tests so that it was easier to test more corners of the find_one behavior, and in the process, I found and fixed a bug in its implementation.

### DIFF
--- a/src/posit/connect/paginator.py
+++ b/src/posit/connect/paginator.py
@@ -33,7 +33,6 @@ class Paginator:
         result = []
         while self.total is None or self.seen < self.total:
             result += self.get_next_page()
-            self.page_number += 1
         return result
 
     def get_next_page(self) -> List[dict]:
@@ -46,4 +45,5 @@ class Paginator:
             self.total = response["total"]
         results = response["results"]
         self.seen += len(results)
+        self.page_number += 1
         return results
--- a/tests/posit/connect/test_users.py
+++ b/tests/posit/connect/test_users.py
@@ -6,7 +6,7 @@ from requests import HTTPError
 
 from posit.connect.client import Client
 
-from .api import load_mock
+from .api import load_mock  # type: ignore
 
 
 class TestUsers:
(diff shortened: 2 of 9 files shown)


tests/posit/connect/test_users.py
```diff
@@ -1,51 +1,25 @@
...
-                        "active_time": "2023-05-05T20:29:11Z",
-                        "confirmed": True,
-                        "locked": False,
-                        "guid": "87c12c08-11cd-4de1-8da3-12a7579c4998",
-                    },
-                ],
-                "current_page": 1,
-                "total": 3,
-            },
+            json=load_mock("v1/users?page_number=1&page_size=2.json"),
```
nealrichardson, line 22: I've made the file paths match the queries in order to make it clear from the file system what they are, and I also aspire to simplifying/encapsulating some of this repeated `responses.get()` invocation further.

src/posit/connect/paginator.py
```diff
@@ -46,4 +45,5 @@ def get_next_page(self) -> List[dict]:
             self.total = response["total"]
         results = response["results"]
         self.seen += len(results)
+        self.page_number += 1
```
jonkeane, line 48: Was this a bug you uncovered using these? The change here seems very sensible, but I'm surprised it was the other way around before.

nealrichardson: Yes, see https://github.com/posit-dev/posit-sdk-py/pull/86/commits/c2b20d8f87a62850140e779dbe7ead6061764ea3. Users.find_one was never incrementing the page number, but the previous test only checked finding something on the first page. After factoring out the test fixtures, I could more easily set up additional tests to confirm different features of this, and encountered this.

PR #94, opened 2024-03-14
feat: switch User and Content to inherit dict, add update method for User

nealrichardson:
To help us move the discussion forward. I switched User to inherit from `dict` to see how that works, including adding an update method and wiring that up. I first tried `UserDict` but it messed up the sort order of the columns in the DataFrame version, and googling a bit, it sounded like it wasn't so bad to inherit from `dict` anymore, so I figured I'd try that and see how it went. Summary of what I found so far, as reflected in the test suite:

* ✅  Forwards and backwards compatibility (i.e., what happens if the API returns more or fewer fields than the class definition knows about) all works fine. If new fields appear and you haven't defined `property`s for them, it doesn't break, and you can still get at them via `[`. Missing fields also are fine, at least as long as you don't access them (you'd get a KeyError).
* ✅ `update()` method works. I was worried from that blog post that you wouldn't be able to define a new one, but it was fine. I did not try setattr, and my guess is that we'd want to disable that anyway (but that's a discussion for another time). 
* 🤔 `pd.DataFrame(list of Users)` works fine since it *is* a dict still, but: in adding the `session` and `url` to the object in order to implement `update()`, those now show up in the data frame. AFAIK there's no way to hide them somewhere else, it's just a dict. 
* 🤔 `@property`s work to access the data, but the type hints don't enforce anything. I didn't add in the datetime parsing, just to see what happens, and the methods happily return the unparsed strings. Not strictly a problem, but worth keeping in mind.
* Other consideration related to the last two: we could add an `__init__` method to do things like parse timestamps. As we discussed before, there are tradeoffs to doing this eagerly vs. on demand. Alternatively, if we kept it lazy (my preference), we could add a `to_pandas` function or something that would parse them when you do that conversion, and leave the attribute parsing to the `property` getters.

### DIFF
--- a/src/posit/connect/users.py
+++ b/src/posit/connect/users.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Any, Callable, List
+from typing import Any, Callable, List, Optional
 
 from requests import Session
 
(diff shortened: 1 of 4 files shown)


nealrichardson:
Ok @tdstein I think this is done (enough) and we should merge so we can unblock other work.

tdstein:
> 🤔 pd.DataFrame(list of Users) works fine since it is a dict still, but: in adding the session and url to the object in order to implement update(), those now show up in the data frame. AFAIK there's no way to hide them somewhere else, it's just a dict.

Can we implement an `__init__` method in Resource as follows? I haven't tried it yet.

```py
def __init__(self, session: requests.Session, url: str, **kwargs):
    self._session = session
    self._url = url
    super().__init__(**kwargs)
```

tdstein:
> Other consideration related to the last two: we could add an __init__ method to do things like parse timestamps. As we discussed before, there are tradeoffs to doing this eagerly vs. on demand. Alternatively, if we kept it lazy (my preference), we could add a to_pandas function or something that would parse them when you do that conversion, and leave the attribute parsing to the property getters.

Let's go ahead and defer this decision until later. I'm satisfied that `pandas.DataFrame(users)` works. 😄

src/posit/connect/users.py
```diff
@@ -13,18 +11,90 @@
...
+            kwargs["username"] = username
+        if first_name is not None:
+            kwargs["first_name"] = first_name
+        if last_name is not None:
+            kwargs["last_name"] = last_name
+        if user_role is not None:
+            kwargs["user_role"] = user_role
+        # if locked is not None:
+        #     kwargs["locked"] = locked
+        self._update(kwargs)
```
tdstein, line 97: This looks good. But should we enforce these logical constraints or defer this behavior to the server? In other words, how do we handle behavioral changes to the PATCH endpoint between versions of Connect?

tdstein (approved):
> Ok @tdstein I think this is done (enough) and we should merge so we can unblock other work.

Agreed. Let's merge and continue iterating.

nealrichardson:
> > 🤔 pd.DataFrame(list of Users) works fine since it is a dict still, but: in adding the session and url to the object in order to implement update(), those now show up in the data frame. AFAIK there's no way to hide them somewhere else, it's just a dict.
> 
> Can we implement an `__init__` method in Resource as follows? I haven't tried it yet.
> 
> ```python
> def __init__(self, session: requests.Session, url: str, **kwargs):
>     self._session = session
>     self._url = url
>     super().__init__(**kwargs)
> ```

I don't think that will succeed in hiding the session and url from the dict itself since it inherits dict, but we can experiment more. Let's merge this and then your other PR, and then we can tinker with the base class.

PR #98, opened 2024-03-15
feat: adds base class 'Resource' and modifies 'User'.

tdstein:


tests/posit/connect/test_resources.py
```diff
@@ -1,8 +1,53 @@
...
+
+class FakeResource(Resource):
+    @property
+    def foo(self) -> str | None:
+        return self.get("foo")
+
+
+class TestResource:
+    def test__getitem__(self):
+        warnings.filterwarnings("ignore", category=FutureWarning)
```
tdstein, line 17: We should move this to a global test suite setup to reduce noise.

nealrichardson: Why? Our tests shouldn't generally be raising this warning, except when we're testing explicitly for it, right?

src/posit/connect/users.py
```diff
@@ -10,21 +10,53 @@
...
+    @property
+    def last_name(self) -> Optional[str]:
+        return self.get("last_name")
+
+    @property
+    def user_role(self) -> Optional[str]:
+        return self.get("user_role")
+
+    @property
+    def created_time(self) -> Optional[datetime]:
```
nealrichardson, line 42: I guess these are just hints for the IDE, they don't seem to enforce anything. This doesn't return datetime in the tests. 

So we should at least switch to `Optional[str]`. But if they're just informative for tab completion or whatever, perhaps we should just say what they're supposed to be, whether or not they're present in a given API response?

```suggestion
    def created_time(self) -> str:
```

src/posit/connect/resources.py
```diff
@@ -1,10 +1,35 @@
...
+            FutureWarning,
+        )
+        return super().__getitem__(key)
+
+    def __setitem__(self, key, value):
+        warnings.warn(
+            f"__setitem__ for '{key}' does not support backwards compatibility. Consider using field based access instead: 'instance.{key} = {value}",
+            FutureWarning,
+        )
+        return super().__setitem__(key, value)
```
nealrichardson, line 23: As we were discussing separately, I'm not sure we want to allow setitem (or delitem, for that matter).

nealrichardson (approved):
A few notes and suggestions. I'm ok to merge whenever you want, I have a few 🤔 about this in practice but we should just push forward (so we can actually get to "in practice" 😂 ) and we'll keep revising as we go.

src/posit/connect/users.py
tdstein: I like the expressive nature of matching the API response schema, but this breaks down between versions of Connect. Therefore, from a software correctness perspective, `Optional` is always correct. 

That logic defines the return type on this method as `Optional[str].`

Do you come to the same conclusion?

tests/posit/connect/test_resources.py
tdstein: Any test which invoke `Resource#__get_item__` will display the warning message. However, the warnings module is designed to only show the warning message once per argument set.

nealrichardson: Right, I just figured that if we were calling getitem in a test for some reason, we should also assert the warning in the test

src/posit/connect/users.py
nealrichardson: Feels a little hyperbolic to think that we don't know what the API will respond in the future. I also figure we'd be adding backwards compat logic if we rename fields. So IMO I'd think we'd want a type hint that was precise and 99% accurate rather than something that was vague and 100% accurate. But then again, I'm probably not the person you want to take advice from on type annotations ;)

tdstein: That's fair. Apologies if the following is obvious...

I tend to think of the type system from a software correctness perspective. So, always returning Optional for these properties would ensure that software written with this version of the SDK will continue to run successfully when a field name changes. This puts the responsibility on the consumer to trigger an error if `created_time` returns None and their application requires a non-None value. Anyone analyzing their code with MyPy will be forced to consider this scenario if we return `Optional` here.

e.g.,
```
created_time = user.created_time
if created_time is None:
    raise RuntimeError

# continue using created_time
```

The flip side returns `str` and then raises an error if the value does not exist. This puts the responsibility on us to keep the SDK up to date and the user to make sure they are using the latest version of the SDK.

```
def created_time(self) -> str:
    if 'created_time' in self:
        return self['created_time']
    raise RuntimeError("'created_time' not found in the response object. Please upgrade to the latest version...)
```

--

After thinking it through, I realized that the second option is a better user experience overall. I want to modify the properties to mimic the expected API behavior and provide actionable error messages.

nealrichardson: IMO we shouldn't let fear of the possibility of changing field names in the API response drive too much of our decisionmaking since we're much more likely to add new fields than remove them, and we'd only remove after a deprecation cycle, which we'd be managing in the SDK. 

I think it's also safe to assume (because we can strongly recommend it) that interactive users are running on the latest version of the SDK, which will know about API changes. If we know that a field only was added in some new version of connect, we can check the Connect version in the `property` and error if it's not new enough. 

The place where we want to be forgiving of changes is with deployed content, we want it to be robust to Connect upgrades. Type hints are irrelevant there because the content isn't being edited. There is a risk of some distant future, after a deprecation cycle, we remove an attribute, and you could have content referencing it with an old SDK version. But, we could also provide a script (using the SDK) that checks package versions in your content before you upgrade Connect and alert you to which ones you need to upgrade the sdk in.

PR #103, opened 2024-03-18
refactor: moves session and url to resource attributes

### LINKED ISSUE #102: Remove `session` and `url` from the underlying hash table in `Resource`.
During `Resource` initialization, the `session` and `url` are stored as items within the underlying hash table. This results in unexpected behavior when a resource list into a pandas DataFrame.

Try to modify the usage of `session` and `url` so that they exist as internal properties instead of on the underlying table.

tdstein:
This change moves the `url` and `session` items from the underlying hash table used by `dict` to instance attributes. 

Subsequently, these columns no longer appear as columns when converting from a `list[Resource]` to a pandas DataFrame.

The `__setattr__` override behavior is also moved to the base `Resource` class.

Resolves #102

### DIFF
--- a/src/posit/connect/resources.py
+++ b/src/posit/connect/resources.py
@@ -1,13 +1,22 @@
 import warnings
 
 from abc import ABC, abstractmethod
-from typing import Generic, List, Optional, TypeVar
+from typing import Any, Generic, List, Optional, TypeVar
+
+import requests
 
 
 T = TypeVar("T")
 
 
 class Resource(ABC, dict):
+    def __init__(self, session: requests.Session, url: str, **kwargs):
+        super().__init__(**kwargs)
+        self.session: requests.Session
+        super().__setattr__("session", session)
+        self.url: str
+        super().__setattr__("url", url)
+
     def __getitem__(self, key):
         warnings.warn(
             f"__getitem__ for '{key}' does not support backwards compatibility. Consider using field based access instead: 'instance.{key}'",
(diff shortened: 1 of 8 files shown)


src/posit/connect/resources.py
```diff
@@ -3,11 +3,18 @@
...
+import requests
+
 
 T = TypeVar("T")
 
 
 class Resource(ABC, dict):
+    def __init__(self, session: requests.Session, url: str, **kwargs):
+        self._session = session
+        self._url = url
```
nealrichardson, line 15: `__setattr__` was blocked because we want ambiguity in whether the object we have locally is in sync with the server, and we should be clear when an action will result in a network request. Routing everything through `.update()` makes that explicit.

I think you can retain the `__setattr__` (which belongs in the `Resource` class, not `User`) if you do `super().__setattr__("session", session)` here.

Unrelated comment to that: can we keep them as `session` and `url` without the `_` prefix? We should be consistent across all of our objects.

tdstein: That works! Thank you!

src/posit/connect/resources.py
```diff
@@ -1,13 +1,22 @@
...
+from typing import Any, Generic, List, Optional, TypeVar
+
+import requests
 
 
 T = TypeVar("T")
 
 
 class Resource(ABC, dict):
+    def __init__(self, session: requests.Session, url: str, **kwargs):
```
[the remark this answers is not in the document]
tdstein: It would muddy the source code since references will then need to be Optional aware

https://github.com/posit-dev/posit-sdk-py/pull/103/files#diff-e5a7adb6609971528ad9aa73e3641f622e37467b94f19bc21c847c9e4cb86aafR61

PR #105, opened 2024-03-19
feat: adds count method to resources

### LINKED ISSUE #47: Collections should have `__len__`
Relates to #35 and #46

tdstein:
Resolves #47

### DIFF
--- a/src/posit/connect/users.py
+++ b/src/posit/connect/users.py
@@ -148,3 +148,10 @@ class Users(Resources[User]):
 
     def delete(self) -> None:
         raise NotImplementedError()
+
+    def count(self) -> int:
+        response = self.session.get(self.url)
+        result: dict = response.json()
+        if "total" not in result:
+            raise RuntimeError("'total' not found in response body")
+        return result.get("total", 0)
--- a/src/posit/connect/users.py
+++ b/src/posit/connect/users.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 from typing import Callable, List, Optional
 
 
-from requests import Session
+import requests
 
 
 from . import urls
@@ -95,7 +95,7 @@ class User(Resource):
 
 
 class Users(Resources[User]):
-    def __init__(self, config: Config, session: Session) -> None:
+    def __init__(self, config: Config, session: requests.Session) -> None:
         self.url = urls.append_path(config.url, "v1/users")
         self.config = config
         self.session = session
@@ -150,8 +150,6 @@ class Users(Resources[User]):
         raise NotImplementedError()
 
     def count(self) -> int:
-        response = self.session.get(self.url)
+        response: requests.Response = self.session.get(self.url, json={"page_size": 1})
         result: dict = response.json()
-        if "total" not in result:
-            raise RuntimeError("'total' not found in response body")
-        return result.get("total", 0)
+        return result["total"]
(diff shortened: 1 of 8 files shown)


src/posit/connect/users.py
```diff
@@ -151,3 +151,10 @@ def update(self) -> User:
 
     def delete(self) -> None:
         raise NotImplementedError()
+
+    def count(self) -> int:
+        response = self.session.get(self.url)
```
nealrichardson, line 156: Pull less data?

```suggestion
        response = self.session.get(self.url, json={"page_size": 1})
```

src/posit/connect/users.py
```diff
@@ -151,3 +151,10 @@ def update(self) -> User:
...
+        result: dict = response.json()
+        if "total" not in result:
+            raise RuntimeError("'total' not found in response body")
+        return result.get("total", 0)
```
nealrichardson, line 160: Can't see that we'd break this API, and even if we did, I think it's fine to let the `KeyError` bubble up.

```suggestion
        return response.json()["total"]
```

tests/posit/connect/__api__/v1/users.json
```diff
@@ -0,0 +1,19 @@
+{
```
nealrichardson, line 1: I'd delete this and just use `tests/posit/connect/__api__/v1/users?page_number=1&page_size=2.json` in the test

tests/posit/connect/test_users.py
```diff
@@ -266,3 +266,13 @@ def test_user_cant_setattr(self):
             match=r"Cannot set attributes: use update\(\) instead",
         ):
             carlos.first_name = "Carlitos"
+
+    @responses.activate
+    def test_count(self):
+        responses.get(
+            "https://connect.example/__api__/v1/users",
+            json=load_mock("v1/users.json"),
+        )
```
nealrichardson, line 275: If you took the suggestion to only request page_size=1 above, add this here:

```
            match=[
                responses.matchers.query_param_matcher(
                    {"page_size": 1}
                )
            ],

File: tinkering.py [hunk 5]
```diff
@@ -1,11 +1,4 @@
...
 with Client() as client:
-    print(client.get("v1/users"))
-    print(client.users.get("f55ca95d-ce52-43ed-b31b-48dc4a07fe13"))
-
-    users = client.users
-    users = users.find(lambda user: user["first_name"].startswith("T"))
-    users = users.find(lambda user: user["last_name"].startswith("S"))
-    user = users.find_one(lambda user: user["user_role"] == "administrator")
-    print(user)
+    print(client.users.count())
```
nealrichardson, line 4: Do you want to commit this? (Or maybe gitignore the file?)

tdstein: It's probably time to delete it. I think we are past the bootstrapping phase

src/posit/connect/users.py
```diff
@@ -148,3 +148,8 @@ def update(self) -> User:
 
     def delete(self) -> None:
         raise NotImplementedError()
+
+    def count(self) -> int:
+        response: requests.Response = self.session.get(self.url, json={"page_size": 1})
```
nealrichardson, line 153: For my own edification: is this type hint necessary? Doesn't it (waves hands) know that `requests.Session.get()` returns `requests.Response`? And that `Response.json()` returns `dict`?

tdstein: No, it isn't necessary. Once I added the correct type declaration to `session` in the `__init__` method, this type declaration became redundant.

Your understanding is correct 😄

PR #116, opened 2024-03-20
refactor: moves pagination logic to paginator

tdstein:
The `page_size` parameters are removed from the Paginator methods. It doesn't seem like we actually need this right now.

And I removed the `lambda` methods from `.find` and `.find_one` since we are ready to start implementing "real" server side filtering. This was always a stop-gap to demonstrate functionality.


src/posit/connect/paginator.py
```diff
@@ -1,49 +1,95 @@
...
+                yield page
+            else:
+                # stop if the result set is empty
+                return
+
+            count += len(page.results)
+            # Check if the local count has reached the total threshold.
+            # It is possible for count to exceed total if the total changes
+            # during execution of this loop.
+            # It is also possible for the total to change between iterations.
```
tdstein, line 78:
```suggestion
            # Check if the local count has reached the total threshold.
            # It is possible for the count to exceed the total if the total changes
            # between loop iterations.
```

tests/posit/connect/test_users.py
```diff
@@ -109,14 +108,14 @@ def test_users_find(self):
             "https://connect.example/__api__/v1/users",
             match=[
                 responses.matchers.query_param_matcher(
-                    {"page_size": 2, "page_number": 2}
+                    {"page_size": 500, "page_number": 2}
                 )
             ],
             json=load_mock("v1/users?page_number=2&page_size=2.json"),
```
nealrichardson, line 114: Do you want to update the file names to reflect the changed query params?

tests/posit/connect/test_users.py
```diff
@@ -152,51 +151,15 @@ def test_users_find_one(self):
...
-        c = con.users.find_one(lambda u: u.first_name == "Carlos", page_size=2)
-        # Can't isinstance(c, User) bc inherits TypedDict (cf. #23)
-        assert c.username == "carlos12"
-
-        # Now test that if not found, it returns None
-        assert (
-            con.users.find_one(lambda u: u.first_name == "Ringo", page_size=2) is None
-        )
-
-    @responses.activate
```
nealrichardson, line 171: Did you mean to delete this test? This still seems useful, no?

tests/posit/connect/test_users.py
tdstein: Thanks! I cleaned this up.

PR #119, opened 2024-03-20
feat: set user attributes on update

### LINKED ISSUE #99: Finish adding .update() to User
#94 adds a start of an update method to User to prove out the underlying object model. Among the remaining items to do that were left off of #94:

- [x] #117
- [ ] Test updating all of the fields
- [ ] Test error handling

### LINKED ISSUE #117: Use the API response to update the local object

tdstein:
Also fixes issues with `client.me` attribute and uses the correct HTTP method

Resolves #117 
Resolves #99

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -50,6 +50,9 @@ class Client:
     def me(self) -> User:
         url = urls.append_path(self.config.url, "v1/user")
         response = self.session.get(url)
+        body = response.json()
+        guid = body["guid"]
+        url = urls.append_path(self.config.url, f"v1/users/{guid}")
         return User(self.session, url, **response.json())
 
     @property
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -51,9 +51,10 @@ class Client:
         url = urls.append_path(self.config.url, "v1/user")
         response = self.session.get(url)
         body = response.json()
+        # Build the canonical URL for the User based on its GUID
         guid = body["guid"]
         url = urls.append_path(self.config.url, f"v1/users/{guid}")
-        return User(self.session, url, **response.json())
+        return User(self.session, url, **body)
 
     @property
     def oauth(self) -> OAuthIntegration:
--- a/src/posit/connect/users.py
+++ b/src/posit/connect/users.py
@@ -58,8 +58,9 @@ class User(Resource):
         return self.get("locked")  # type: ignore
 
     def _update(self, body):
-        response = self.session.patch(self.url, json=body)
-        # If the request is successful, update the local object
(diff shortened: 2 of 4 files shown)


src/posit/connect/users.py
```diff
@@ -58,12 +58,10 @@ def locked(self) -> bool:
         return self.get("locked")  # type: ignore
 
     def _update(self, body):
-        self.session.patch(self.url, json=body)
-        # If the request is successful, update the local object
-        super().update(body)
-        # TODO(#99): that patch request returns a payload on success,
-        # so we should instead update the local object with that payload
-        # (includes updated_time)
+        if len(body) == 0:
```
nealrichardson, line 61: Should we test this case?

src/posit/connect/client.py
```diff
@@ -50,6 +50,9 @@ def connect_version(self):
     def me(self) -> User:
         url = urls.append_path(self.config.url, "v1/user")
         response = self.session.get(url)
+        body = response.json()
+        guid = body["guid"]
+        url = urls.append_path(self.config.url, f"v1/users/{guid}")
```
nealrichardson, line 55: Is this necessary because you can't `PUT v1/users/`? Should we add a test covering this behavior?

src/posit/connect/client.py
```diff
@@ -50,6 +50,9 @@ def connect_version(self):
     def me(self) -> User:
         url = urls.append_path(self.config.url, "v1/user")
         response = self.session.get(url)
+        body = response.json()
+        guid = body["guid"]
```
nealrichardson, line 54:
```suggestion
        # Build the canonical URL for the User based on its GUID
        guid = body["guid"]
```

Side note: it would be nice if the API didn't require us to construct this URL, either by doing some kind of redirect to the `v1/users/guid/` location or returning it in a Location header or something.

src/posit/connect/client.py
```diff
@@ -50,6 +50,9 @@ def connect_version(self):
...
+        url = urls.append_path(self.config.url, f"v1/users/{guid}")
         return User(self.session, url, **response.json())
```
nealrichardson, line 56:
```suggestion
        return User(self.session, url, **body)
```

nealrichardson (approved):
A couple of suggestions but otherwise LGTM

src/posit/connect/client.py
tdstein: Correct. Yes, we should add some test coverage for this endpoint.

src/posit/connect/client.py
tdstein: Agreed. Some minimal link relations in the response object would go a long way.

PR #120, opened 2024-03-20
feat: adds update support to content

### LINKED ISSUE #89: Add .update() method to Content
I think we need to do this as part of our first release because it will force us to address some basic issues, including:

* Can't inherit from TypedDict (#23)
* Passing around `session` everywhere. We don't currently pass it to `User()` or `ContentItem()`, perhaps we should just do that? Or should there be some global or other means of finding "the" session that corresponds to the URL of the resource you're holding?

tdstein:
Resolves #89 

Here's a recording that demonstrates the type hinting behavior in VSCode.

https://github.com/posit-dev/posit-sdk-py/assets/6015574/df0424c0-fefd-445a-8c74-6a26cdd95fa0

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -11,37 +11,6 @@ from .config import Config
 from .resources import Resources, Resource
 
 
-_UPDATE_FIELDS = frozenset(
-    {
-        "name",
-        "title",
-        "description",
-        "access_type",
-        "owner_guid",
-        "connection_timeout",
-        "read_timeout",
-        "init_timeout",
-        "idle_timeout",
-        "max_processes",
-        "min_processes",
-        "max_conns_per_process",
-        "load_factor",
-        "cpu_request",
-        "cpu_limit",
-        "memory_request",
-        "memory_limit",
-        "amd_gpu_limit",
-        "nvidia_gpu_limit",
-        "run_as",
-        "run_as_current_user",
-        "default_image_name",
-        "default_r_environment_management",
-        "default_py_environment_management",
-        "service_account_name",
-    }
-)
-
-
 class ContentItem(Resource):
     @property
     def guid(self) -> str:
@@ -245,7 +214,7 @@ class ContentItem(Resource):
         service_account_name: Optional[str] = ...,
     ) -> None:
         """
-        Update the content with the specified parameters.
+        Update the content item.
 
         Args:
             name (str): The name of the content.
@@ -275,7 +244,7 @@ class ContentItem(Resource):
             service_account_name (Optional[str]): The service account name.
 
         Returns:
-            None: This method does not return anything.
+            None
         """
         ...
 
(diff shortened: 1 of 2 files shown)


src/posit/connect/content.py
```diff
@@ -183,6 +183,94 @@ def app_role(self) -> str:
...
+        memory_request: Optional[int] = None,
+        memory_limit: Optional[int] = None,
+        amd_gpu_limit: Optional[int] = None,
+        nvidia_gpu_limit: Optional[int] = None,
+        run_as: Optional[str] = None,
+        run_as_current_user: Optional[bool] = None,
+        default_image_name: Optional[str] = None,
+        default_r_environment_management: Optional[bool] = None,
+        default_py_environment_management: Optional[bool] = None,
+        service_account_name: Optional[str] = None,
```
mmarchetti, line 212: For forward compatibility, would it make sense to accept `**kw` here in addition to the known/typed attributes? The behavior could be to issue a warning (in case it's a simple typo) and merge the values into the payload. If this makes the payload invalid, Connect will return a 400.

nealrichardson: We could do that. You still get the benefit of tab completion and type hinting for the known parameters, and like you say, if the thing you send is invalid, the server will reject it.

src/posit/connect/content.py
```diff
@@ -183,6 +183,94 @@ def app_role(self) -> str:
...
+            data["read_timeout"] = read_timeout
+        if init_timeout is not None:
+            data["init_timeout"] = init_timeout
+        if idle_timeout is not None:
+            data["idle_timeout"] = idle_timeout
+        if max_processes is not None:
+            data["max_processes"] = max_processes
+        if min_processes is not None:
+            data["min_processes"] = min_processes
+        if max_conns_per_process is not None:
```
nealrichardson, line 237: I think for many of these, we can't check `is not None` because `None` is a valid value in the API, many of these things are int or null. So maybe the check is to see if it is not equal to `self.param`?

tests/posit/connect/test_content.py
```diff
@@ -6,6 +6,28 @@
...
+        responses.get(
+            f"https://connect.example/__api__/v1/content/{guid}",
+            json=load_mock(f"v1/content/{guid}.json"),
+        )
+        con = Client("12345", "https://connect.example")
+        content = con.content.get(guid)
+        assert content.guid == guid
+
+        new_name = "New Name"
+        responses.patch(
```
nealrichardson, line 22: So this is PATCH but user is PUT? 🤦

tdstein: Yep!

nealrichardson: Good thing we're writing an SDK to encapsulate that!

src/posit/connect/content.py
tdstein: I figured out a solution using the Ellipsis literal. Working on a full implementation now.

```py
def update(  # type: ignore
        self,
        name: Union[str, EllipsisType] = ...,
        title: Union[Optional[str], EllipsisType] = ...,
    ) -> None:
        kwargs = {}
        if name is not ...:
            kwargs["name"] = name
        if title is not ...:
            kwargs["title"] = title

        response = self.session.patch(self.url, json=kwargs)
        super().update(**response.json())
```

tdstein: I pushed a new commit that uses an override signature to provide the type hinting. The implementation strictly checks if the provided fields are part of the known request body. But we can modify this logic to support forward compatibility.

tdstein:
Converting back to draft until implementation and testing are complete.

src/posit/connect/content.py
tdstein: Added support for this in the latest iteration.

mmarchetti (approved):
This approach looks interesting to me. It preserves typchecking/hints to the extent possible while allowing for forward compatibility (I was assuming that's a design goal; if not then we can reconsider).

tdstein:
> This approach looks interesting to me. It preserves typchecking/hints to the extent possible while allowing for forward compatibility (I was assuming that's a design goal; if not then we can reconsider).

Thanks! Here's another behavioral artifact that I find interesting. You can define additional overload methods, and the language server will navigate those type hints correctly. So, the method signature effectively becomes the request schema for different API versions. 


https://github.com/posit-dev/posit-sdk-py/assets/6015574/29c0f2c7-6242-4030-8c97-2ae51a907980

PR #122, opened 2024-03-22
refactor: consistent update method signature

tdstein:


tdstein:
@nealrichardson / @mmarchetti - here is a follow-up to #89, which uses the same pattern for the update method.

PR #124, opened 2024-03-22
refactor: inject config instead of url into resource

tdstein:


tdstein:
I'll propogate these changes to my open pull requests once merged.

PR #129, opened 2024-03-22
feat!: removes filter parameter from content.find and content.find_one.

tdstein:
BREAKING CHANGE: Support for passing a filter as lambda expression to `content.find` and `content.find_one` is removed. Please use client-side filtering instead. Additional filter support is planned for a future release.

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Callable, List, Optional, overload
+from typing import List, Optional, overload
 
 
 from requests import Session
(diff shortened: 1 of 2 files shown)


mmarchetti:
What's the motivation for this change? These definitions of `find` and especially `find_one` seem a lot less useful.

tdstein:
Moving towards kwargs with proper server-side filtering based on API support. I am going to try following a similar pattern to the update methods.

tdstein:
@mmarchetti - here's the wip: https://github.com/posit-dev/posit-sdk-py/pull/130/files

PR #130, opened 2024-03-22
feat: adds query params to find and find_one

tdstein:
This change follows the same `@overload` pattern in `user.update`.


tests/posit/connect/test_users.py
```diff
@@ -354,9 +266,123 @@ def test_user_cant_setattr(self):
...
+            "https://connect.example/__api__/v1/users",
+            match=[
+                responses.matchers.query_param_matcher(
+                    {"page_size": 500, "page_number": 1, **params}
+                )
+            ],
+            json=load_mock("v1/users?page_number=1&page_size=500.jsonc"),
+        )
+        con = Client(api_key="12345", url="https://connect.example/")
+        con.users.find_one(**params)
```
mmarchetti, line 304: Does this need an assertion?

tests/posit/connect/test_users.py
```diff
@@ -354,9 +266,123 @@ def test_user_cant_setattr(self):
...
+            "https://connect.example/__api__/v1/users",
+            match=[
+                responses.matchers.query_param_matcher(
+                    {"page_size": 500, "page_number": 2, **params}
+                )
+            ],
+            json=load_mock("v1/users?page_number=2&page_size=500.jsonc"),
+        )
+        con = Client(api_key="12345", url="https://connect.example/")
+        con.users.find(**params)
```
mmarchetti, line 380: Same question here.

mmarchetti (approved):

tests/posit/connect/test_users.py
tdstein: The `match` value above is asserted by `responses`. But, on second thought, an additional assertion on `calls == 1` is wise. I'll add that.

mmarchetti (approved):

PR #145, opened 2024-04-02
feat: adds permissions find, find_one, and update

### LINKED ISSUE #101: Content permissions API wrapper
```
con.users.get(id).permissions.find()

for perm in con.users.get(id).permissions.find():
    perm.update(role="viewer")
```

https://docs.posit.co/connect/api/#put-/v1/content/-guid-/permissions/-id-

tdstein:
Resolves #101


src/posit/connect/resources.py
```diff
@@ -44,31 +44,7 @@ def __setattr__(self, name: str, value: Any) -> None:
...
-    def get(self, *args, **kwargs) -> T:
-        raise NotImplementedError()
-
-    @abstractmethod
-    def update(self, *args, **kwargs) -> T:
-        raise NotImplementedError()
-
-    @abstractmethod
-    def count(self, *args, **kwargs) -> int:
-        raise NotImplementedError()
```
tdstein, line 74: I can cherry-pick this out if needed. I'm realizing that not all API endpoints can support every one of these methods.

nealrichardson: Fine by me to leave it here (and I'm always happy about deleting code 😂 )

src/posit/connect/permissions.py
```diff
@@ -0,0 +1,107 @@
...
+
+        The conventional name for this field is 'guid'. As of v2024.03.0, the API uses the field name 'id'.
+
+        Returns
+        -------
+        str
+        """
+        # Alias 'id' to 'guid'.
+        # The conventional name for this field across applications is 'guid'.
+        return self.get("id", self.get("guid"))  # type: ignore
```
nealrichardson, line 28: *Is* the id a GUID? Doesn't look like it according to the API docs, but the example there could be wrong.

Should we file an issue on Connect to correct this inconsistency?

tests/posit/connect/test_permissions.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class TestPermissionsFind:
+    @responses.activate
+    def test(self):
+        # test data
+        content_guid = str(uuid.uuid4())
+        fake_permissions = [
+            {
+                "guid": str(uuid.uuid4()),
```
nealrichardson, line 122: But isn't the field called "id" in the API?

Also, can we put these JSON API responses in a file like the others?

src/posit/connect/permissions.py
tdstein: Ah, good catch. I just assumed it was a GUID. 👍🏻 

It's probably a discussion for another time, but I'm surprised to see that it's sequential. 

side note; I noticed that the request body schema definition here is incorrect. The example is correct, but the schema definition is missing the required `principal` fields. I'll file a ticket for it once I wrap things up.

tdstein: It would help if I provided the link: https://docs.posit.co/connect/api/#put-/v1/content/-guid-/permissions/-id-

PR #147, opened 2024-04-02
feat: adds permissions create

tdstein:
Related to #101

### DIFF
--- a/tests/posit/connect/test_permissions.py
+++ b/tests/posit/connect/test_permissions.py
@@ -1,3 +1,4 @@
+import random
 import uuid
 
 from unittest.mock import Mock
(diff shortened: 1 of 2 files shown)


tests/posit/connect/test_permissions.py
```diff
@@ -106,6 +106,51 @@ def test_role_update(self):
...
+class TestPermissionsCreate:
+    @responses.activate
+    def test(self):
+        # data
+        id = "94"
+        content_guid = "f2f37341-e21d-3d80-c698-a935ad614066"
+        principal_guid = str(uuid.uuid4())
+        principal_type = "user"
+        role = "owner"
+        fake_permission = {
```
nealrichardson, line 118: This is an odd construction to my (👴) eyes, would this be the same as `load_mock(...).update({stuff})`?

tdstein: Ha. It is. This style is muscle memory from my TypeScript days.

e.g.,
```ts
o1 = { 'a': 'b' }
o2 = { ...o1, 'c': 'd' }
```

Do you know if either style is more conventional in modern Python?

nealrichardson: To me, `dict.update()` makes it clear that you're starting with a dict and adding/overwriting some of the values, so I'd go with that personally.

nealrichardson: It's NBD of course. I just like readable/well commented test code so that it's clear what's happening to someone who may only be a new or casual contributor, would rather it be explicit and verbose and not overly clever.

tdstein: Yeah, I'm glad you asked. I was reading up on PEP-584, which goes into the history of merging dictionaries.

https://peps.python.org/pep-0584/#motivation

Support for union operators was added in Python 3.10. It appears this is the preferred one-liner. I suppose this is a bit cleaner once you are familiar with it: `d2 = d1 | { 'c': 'd' }`

tdstein: Alas, there is no backport to Python 3.8 or 3.9. So, I'm going to stick with creating and updating for now.

PR #152, opened 2024-04-02
feat!: renames `Client` to `Connect`

tdstein:
BREAKING CHANGE: `posit.connect.Client` has been renamed as `posit.Connect`. Please replace all references to `Client` with `Connect` when upgrading.

Previous:

```

c = Client()
```

Current:

```
from posit import Connect

c = Connect()
```

### DIFF
--- a/examples/0001-overview.qmd
+++ b/examples/0001-overview.qmd
@@ -7,41 +7,41 @@ The examples cover various aspects, including authentication, data retrieval, an
 ## Example - Print information for each user where `user_role='publisher'`
 
 ```python
-from posit.connect import Client
+from posit.connect import Connect
 
-with Client() as client:
-    for user in client.users.find(user_role='publisher'):
+with Connect() as connect:
+    for user in connect.users.find(user_role='publisher'):
         print(user)
 ```
 
 ## Example - Print information for a single user where `prefix='b'`
 
 ```python
-from posit.connect import Client
+from posit.connect import Connect
 
-with Client() as client:
-    user = client.users.find_one(prefix='b')
+with Connect() as connect:
+    user = connect.users.find_one(prefix='b')
     print(user)
 ```
 
 ## Example - Print the title for each content item that I own.
 
 ```python
-from posit.connect import Client
+from posit.connect import Connect
 
-with Client() as client:
-    guid = client.me.guid
-    for item in client.content.find(owner_guid=guid)
+with Connect() as connect:
+    guid = connect.me.guid
+    for item in connect.content.find(owner_guid=guid)
         print(item.title)
 ```
 
 ## Example - Update the title for a content item.
 
 ```python
-from posit.connect import Client
+from posit.connect import Connect
 
-with Client() as client:
+with Connect() as connect:
     guid = ... # the guid for a content item
-    item = client.content.get(guid)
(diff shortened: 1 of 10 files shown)


tdstein:
I plan to do a release prior to and directly after this change is merged. All existing work in `main` will be released under `0.2.0`. This change will be released as `0.3.0`.

src/posit/__init__.py
```diff
@@ -1 +1,3 @@
 """The Posit SDK."""
+
+from .connect import Connect  # noqa
```
nealrichardson, line 3: Is it confusing that both `from posit import connect` and `from posit import Connect` are correct but do very different things? 

Even if it is a good idea (not making a judgment on that), it will probably cause hell for generating documentation on case-insensitive file systems (macOS) because we'll have man pages for `posit.connect` and `posit.Connect` colliding with each other. On Arrow for example, I could basically never touch the sphinx docs because of collisions like this.

examples/0001-overview.qmd
```diff
@@ -7,41 +7,41 @@ The examples cover various aspects, including authentication, data retrieval, an
 ## Example - Print information for each user where `user_role='publisher'`
 
 ```python
-from posit.connect import Client
+from posit.connect import Connect
 
-with Client() as client:
-    for user in client.users.find(user_role='publisher'):
+with Connect() as connect:
```
nealrichardson, line 12: maybe `with Connect() as con` or something, just to disambiguate? (Or is that more ambiguous?)

src/posit/__init__.py
tdstein: That's an interesting edge case. I haven't run into it before, but it makes perfect sense.

After doing some more digging, I noticed that the pattern we currently have seems common (Google, AWS). The only difference I see is that in SDKs where `Client` is used without a service name prefix, the reference documentation typically follows this pattern:

(aside; it's also common in other languages that promote namespacing via packages. Java and Go, for example.)

```py
from posit import connect

client = connect.Client()
```

The other common approach (Stripe, Mongo, Azure) is prefixing the product name (i.e., `ConnectClient`).

```py
from posit import ConnectClient

client = ConnectClient()
```

I personally like the first/current option more since it provides a generic namespace for any other "Connect" specific code. However, the second option is also quite nice due to its simplicity and verbosity. If I had to bet, I would assume the popular vote would be for `ConnectClient`. Therefore, I think our best option is to do both via:

`src/posit/__init__.py`
```py
from posit.connect import Client as ConnectClient
```

tdstein: This will extend nicely to other products as well.

```py
from posit.connect import Client as ConnectClient
from posit.package_manager import Client as PackageManagerClient
from posit.workbench import Client as WorkbenchClient
```

mmarchetti: My preference would be to have `one-- and preferably only one --obvious way to do it`.
```
client = Client()
```

seems idiomatic to me, as does

```
from posit import connect
client = connect.Client()
```

I would probably only use the latter if I was planning on talking to multiple Posit products in the same script, or if I needed something else from the `connect` namespace.

tdstein:
Closing based on discussion feedback. What we currently have will work.

Issue #140, opened 2024-04-01
Better docstrings for content.update


nealrichardson:
They're all pretty tautological, just restating the field name. 

```
            name (str): The name of the content.
            title (Optional[str]): The title of the content.
            description (str): The description of the content.
            access_type (str): The access type of the content.
...
```

There's more useful information in the [API docs](https://docs.posit.co/connect/api/#patch-/v1/content/-guid-), explaining what things are for, what valid values are, etc. We should grab those from the swagger spec and put them in there, translating to python as appropriate. 

Probably also would be good to include xrefs to the hosted connect API docs for the relevant APIs.

tdstein:
I've been going back and forth on whether to link directly to the API documentation. It's possible for the hosted documentation to diverge from the docstrings over time.

tdstein:
Related to #135

PR #155, opened 2024-04-03
feat: adds content create

### LINKED ISSUE #106: Content.create()
https://docs.posit.co/connect/api/#post-/v1/content

tdstein:
Resolves #106

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -18,6 +18,10 @@ class ContentItem(Resource):
     """A piece of content."""
 
     @property
+    def id(self) -> str:
+        return self.get("id")  # type: ignore
+
+    @property
     def guid(self) -> str:
         return self.get("guid")  # type: ignore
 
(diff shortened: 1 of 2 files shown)


tdstein:
I'll address the docstrings as part of #140

src/posit/connect/content.py
```diff
@@ -17,6 +17,10 @@
 class ContentItem(Resource):
     """A piece of content."""
 
+    @property
+    def id(self) -> str:
```
nealrichardson, line 21: Only looking at this since you moved it: should we actually remove this accessor? Is my understanding correct that `id` is effectively deprecated and that `guid` is the real identifier you should always use? @aronatkins?

aronatkins: We originally did not include `id` in the public v1 API. The `id` was added afterwards in order to help with cross-referencing against the on-disk layout and server logs (which did not include guid at that time).

At this point, it is probably too strong a statement to declare `id` as deprecated, but it cannot be used to look-up content items using the v1 API.

PR #159, opened 2024-04-08
fix: removes resource warning messages.

tdstein:
IPython utilizes custom i/o handlers when writing to stdout and stderr. During normal usages (.e.g, `client.users.find()`), these handlers invoke __getitem__ resulting in a warning messages for each key in the resource.

This conclusion is based on observed behavior using IPython. I haven't determined the root cause of the issue. Another fix could be checking if __getitem__, __setitem__, and __delitem__ are invoked in an IPython process.

### DIFF
--- a/src/posit/connect/resources.py
+++ b/src/posit/connect/resources.py
@@ -19,27 +19,6 @@ class Resource(ABC, dict):
         self.session: requests.Session
         super().__setattr__("session", session)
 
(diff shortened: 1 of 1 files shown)


nealrichardson (approved):
LGTM. I think we can safely discourage this access pattern by never doing it in our examples, warnings won't be needed.

PR #160, opened 2024-04-08
feat: adds owner property to content-item.

### LINKED ISSUE #158: Add `owner` details to the output of `connect.content.find()`
for consistency with R `connectapi` package, the output of `connect.content.find()` should include an owner column with details of the content item owner(`owner$guid`, `owner$first_name`, `owner$last_name` and `owner$username`). 
This would enable any audit functionality we are building in the cookbooks as we would be able to get both content and owner details in one function(something which can already be done in R).

tdstein:
Resolves #158

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -205,8 +205,8 @@ class ContentItem(Resource):
     # CRUD Methods
 
     @property
-    def tags(self) -> List[dict] | None:
-        return self.get("tags")
+    def tags(self) -> List[dict]:
+        return self.get("tags", [])
 
     def delete(self) -> None:
         """Delete the content item."""
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -191,6 +191,10 @@ class ContentItem(Resource):
         return self.get("owner_guid")  # type: ignore
 
     @property
+    def owner(self) -> str:
+        return self.get("owner", {})  # type: ignore
+
+    @property
     def content_url(self) -> str:
         return self.get("content_url")  # type: ignore
 
@@ -202,11 +206,11 @@ class ContentItem(Resource):
     def app_role(self) -> str:
         return self.get("app_role")  # type: ignore
 
-    # CRUD Methods
-
     @property
     def tags(self) -> List[dict]:
         return self.get("tags", [])
+
+    # CRUD Methods
 
     def delete(self) -> None:
         """Delete the content item."""
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -392,7 +392,10 @@ class Content(Resources):
 
     @overload
     def find(
-        self, owner_guid: str = ..., name: str = ..., include: str = "owner,tags"
+        self,
+        owner_guid: str = ...,
+        name: str = ...,
+        include: Optional[str] = "owner,tags",
     ) -> List[ContentItem]:
         """Find content items.
 
@@ -402,7 +405,7 @@ class Content(Resources):
             The owner's unique identifier, by default ...
         name : str, optional
             The simple URL friendly name, by default ...
-        include : str, optional
+        include : Optional[str], optional
             Comma separated list of details to include in the response, allows 'owner' and 'tags', by default 'owner,tags'
 
         Returns
@@ -412,20 +415,37 @@ class Content(Resources):
         ...
 
     @overload
-    def find(self, *args, **kwargs) -> List[ContentItem]:
-        ...
-
-    def find(self, *args, **kwargs) -> List[ContentItem]:
+    def find(
+        self, *args, include: Optional[str] = "owner,tags", **kwargs
+    ) -> List[ContentItem]:
         """Find content items.
+
+        Parameters
+        ----------
+        include : Optional[str], optional
+            Comma separated list of details to include in the response, allows 'owner' and 'tags', by default 'owner,tags'
 
         Returns
         -------
         List[ContentItem]
         """
-        params = dict(*args, **kwargs)
-        if "include" not in params:
-            params["include"] = "owner,tags"
+        ...
 
+    def find(
+        self, *args, include: Optional[str] = "owner,tags", **kwargs
+    ) -> List[ContentItem]:
+        """Find content items.
+
+        Parameters
+        ----------
+        include : Optional[str], optional
+            Comma separated list of details to include in the response, allows 'owner' and 'tags', by default 'owner,tags'
+
+        Returns
+        -------
+        List[ContentItem]
+        """
+        params = dict(*args, include=include, **kwargs)
         response = self.session.get(self.url, params=params)
         results = response.json()
         items = (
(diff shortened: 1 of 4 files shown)


tests/posit/connect/__api__/v1/content/f2f37341-e21d-3d80-c698-a935ad614066.json
```diff
@@ -37,9 +37,24 @@
...
-    "owner_guid": "87c12c08-11cd-4de1-8da3-12a7579c4998",
     "content_url": "https://connect.example/content/f2f37341-e21d-3d80-c698-a935ad614066/",
     "dashboard_url": "https://connect.example/connect/#/apps/f2f37341-e21d-3d80-c698-a935ad614066",
     "app_role": "viewer",
-    "id": "8274"
-}
\ No newline at end of file
+    "id": "8274",
+    "owner_guid": "20a79ce3-6e87-4522-9faf-be24228800a4",
+    "owner": {
```
nealrichardson, line 45: Interestingly, the `GET v1/content/{guid}/` endpoint does *not* accept owner,tags as a query param, so this would not show up here: https://docs.posit.co/connect/api/#get-/v1/content/-guid-

Other fun detail I discovered on Friday: in `GET v1/content/?include=tags,owner`, while `owner` is always populated, `tags` is only added if there are any tags associated with the content; I expected to see `tags: []` but did not see that when inspecting real responses from Connect. 

I'd recommend copying the contents of the JSON from here: https://github.com/rstudio/connectapi/pull/223/files#diff-0ff270391c558938645b02da8e606b3e9bb14a131bf4d90a6ad21a2541871712 I was adding the API mocks we have in posit-sdk-py over there and had to add tags and owners, so that JSON there is consistent with the other mocks we have here, and it has some variation in whether or not tags exist.

src/posit/connect/content.py
```diff
@@ -189,10 +189,18 @@ def dashboard_url(self) -> str:
     def app_role(self) -> str:
         return self.get("app_role")  # type: ignore
 
+    @property
+    def owner(self) -> dict | None:
+        return self.get("owner")
```
nealrichardson, line 194: Since `GET v1/content/{guid}/` does not allow you to include owner or tags, this should first check for the existence of `owner` in the data, and if not found, `GET` the `owner_guid`. 

Should this return a `User` instead of `dict`?

It should never return `None`.

src/posit/connect/content.py
```diff
@@ -189,10 +189,18 @@ def dashboard_url(self) -> str:
...
+
     @property
     def permissions(self) -> Permissions:
         return Permissions(self.config, self.session, content_guid=self.guid)
 
+    @property
+    def tags(self) -> List[dict] | None:
```
nealrichardson, line 201: This one is trickier since even if you've said to include `tags`, it may not be present in the payload. So if you do the same as I suggested with `owner`, you'll always make an extra request when there are no tags. 

One option could be that when `find()` creates the `ContentItem`s, if `tags` is in the `include` argument and `tag not in result`, to add `tag = []` to the item. That way, `tags` is always a list, and if it is `None` in the data object at this point, that means we didn't request `tags` in the search query, so do the extra `GET` to fetch them. Does that make sense?

nealrichardson (changes requested):
Seems like this should be simple but alas. LMK if my comments don't make sense

src/posit/connect/content.py
tdstein: Should this return a User instead of dict?

The information here is a subset of fields defined in `User`. I considered defining an `Owner(Resource)` instead.

nealrichardson: > Should this return a User instead of dict?

I don't know, that was my question! 🤪 

`dict` is probably fine until we determine we need something richer (and then `User` is just a subclass of `dict` so it shouldn't be a breaking change if we were to upgrade it). The `dict` does include the `guid` so you could fetch the whole `User` if you wanted.

tdstein: lol, I forgot to quote it... 🤦🏻

src/posit/connect/content.py
tdstein: It does, but I'm not wary of getting too fancy with this. This seems like behavior we should modify server side. For the `GET` case, let's modify the behavior to mimic what we "want" the sever to do. I think that is implementing the `include` querystring parameter.

For now, I modified the property to return an empty array by default. This seems logical since the default querystring parameters are set to include tags. If the user were to modify the include options, then they should be aware that tags will be empty.

src/posit/connect/content.py
```diff
@@ -189,10 +189,18 @@ def dashboard_url(self) -> str:
...
+    def owner(self) -> dict | None:
+        return self.get("owner")
+
     @property
     def permissions(self) -> Permissions:
         return Permissions(self.config, self.session, content_guid=self.guid)
 
+    @property
+    def tags(self) -> List[dict]:
+        return self.get("tags", [])
```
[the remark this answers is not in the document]
tdstein: I was looking into fetching tags independently as part of `client.content.get`.

But this is introducing an edge case I hadn't previously considered. We use `@property` to expose dict fields and relationships to other `Resources` (.e.g., content_item.permissions). In this case, `tags` is in conflict because it should be used for both: the dict field and the relationship. 

Let's consider `client.content.get(guid).tags.create('red')`. We could extend the current implementation of `Resources` to allow for this by having `Resources` implement `List[Resource]`. I did a little experimentation and found that it's fairly straightforward. Just requires a bit of modification to how we are using `count`.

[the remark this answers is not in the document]
tdstein: Eh, scratch that. Not straightforward...

src/posit/connect/content.py
```diff
@@ -189,10 +189,18 @@ def dashboard_url(self) -> str:
...
+    @property
+    def owner(self) -> dict:
+        return self.get("owner", {})
+
     @property
     def permissions(self) -> Permissions:
         return Permissions(self.config, self.session, content_guid=self.guid)
 
+    @property
+    def tags(self) -> List[dict]:
```
nealrichardson, line 201: Followup issue for this property: https://github.com/posit-dev/posit-sdk-py/issues/164

src/posit/connect/content.py
```diff
@@ -189,10 +189,18 @@ def dashboard_url(self) -> str:
     def app_role(self) -> str:
         return self.get("app_role")  # type: ignore
 
+    @property
+    def owner(self) -> dict:
+        return self.get("owner", {})
```
nealrichardson, line 194: Followup: https://github.com/posit-dev/posit-sdk-py/issues/165

nealrichardson (commented):
Made some followup issues and a few suggestions on cleaning up the tests but otherwise I think this is fine for now.

tests/posit/connect/test_content.py
```diff
@@ -230,36 +236,123 @@ def test(self):
...
+
+        # setup
+        client = Client("12345", "https://connect.example")
+
+        # invoke
+        content_item = client.content.find_one()
+
+        #  assert
+        assert mock_get.call_count == 1
+        assert content_item
```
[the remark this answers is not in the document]
tdstein: That ends up testing the mocks' wiring, but it can be good to add.

nealrichardson (approved):

PR #161, opened 2024-04-08
feat: adds bundles

### LINKED ISSUE #109: ContentItem.bundles
https://docs.posit.co/connect/api/#get-/v1/content/-guid-/bundles

tdstein:
Adds support for bundles.find, bundles.find_one, bundles.get, and bundle.delete.

Resolves #109

### DIFF
--- a/src/posit/connect/bundles.py
+++ b/src/posit/connect/bundles.py
@@ -90,7 +90,7 @@ class Bundle(Resource):
     def metadata(self) -> BundleMetadata:
         return BundleMetadata(self.config, self.session, **self.get("metadata", {}))
 
-    # Accessor Methods
+    # CRUD Methods
 
     def delete(self) -> None:
         path = f"v1/content/{self.content_guid}/bundles/{self.id}"
--- a/src/posit/connect/bundles.py
+++ b/src/posit/connect/bundles.py
@@ -1,0 +1,129 @@
+from __future__ import annotations
+
+from typing import List
+
+from requests.sessions import Session as Session
+
+from posit.connect.config import Config
+
+from . import urls
+
+from .resources import Resources, Resource
+
+
+class BundleMetadata(Resource):
+    @property
+    def source(self) -> str | None:
+        return self.get("source")
+
+    @property
+    def source_repo(self) -> str | None:
+        return self.get("source_repo")
+
+    @property
+    def source_branch(self) -> str | None:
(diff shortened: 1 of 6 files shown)


src/posit/connect/bundles.py
```diff
@@ -0,0 +1,129 @@
...
+from requests.sessions import Session as Session
+
+from posit.connect.config import Config
+
+from . import urls
+
+from .resources import Resources, Resource
+
+
+class BundleMetadata(Resource):
```
nealrichardson, line 14: Does this need to be a Class? It's not a Resource properly, is it? There's nothing you can do with it. It just feels like a nested field in `Bundle` to me, `Bundle.metadata` is a dict.

src/posit/connect/bundles.py
```diff
@@ -0,0 +1,129 @@
...
+        return [
+            Bundle(
+                self.config,
+                self.session,
+                **result,
+            )
+            for result in results
+        ]
+
+    def find_one(self) -> Bundle | None:
```
nealrichardson, line 120: `find_one()` is pretty pointless without any args or query params, no?

src/posit/connect/content.py
```diff
@@ -189,9 +202,7 @@ def dashboard_url(self) -> str:
     def app_role(self) -> str:
         return self.get("app_role")  # type: ignore
 
-    @property
-    def permissions(self) -> Permissions:
-        return Permissions(self.config, self.session, content_guid=self.guid)
+    # Accessor Methods
```
nealrichardson, line 205: Is `delete` an accessor method?

nealrichardson (commented):
A couple of questions but generally makes sense

src/posit/connect/content.py
tdstein: Since the list of `@property` methods was getting long, I felt the need to organize it. I've used "accessor" to describe things that access external data sources over the years. But maybe "CRUD Methods" is more suitable here?

 ```suggestion
     # CRUD Methods
```

src/posit/connect/bundles.py
tdstein: We have defined `find` and `fine_one` in each of the `Resources` classes thus far, it seems ok to continue that pattern. I'm still up for adding `fetch` and `fetch_one` in place of `find` and `find_one` when there aren't any supported query params for the endpoint.

Issue #137, opened 2024-03-29
Content usage wrapper


jonkeane:
It would be nice to have a wrapper that lets me get content usage _regardless of the content is Shiny or not_. Currently we have [`/v1/instrumentation/shiny/usage`](https://docs.posit.co/connect/api/#get-/v1/instrumentation/shiny/usage) for shiny usage and [`/v1/instrumentation/content/visits`](https://docs.posit.co/connect/api/#get-/v1/instrumentation/content/visits) for everything else. They contain much of the same information (though there isn't complete overlap!), but it looks like enough that we could return a combined list that has the fields when they are relevant and either doesn't have them or they are marked as `NULL` when they are not relevant.

See: https://github.com/rstudio/connectapi/issues/218

nealrichardson:
I believe this will involve https://github.com/posit-dev/posit-sdk-py/issues/75.

PR #163, opened 2024-04-10
feat: adds visits find and find_one

### LINKED ISSUE #75: Add pagination support for keyset (cursor) types
The existing Paginator implementation only supports offset pagination. Implement a second class that supports cursors. 

See https://docs.posit.co/connect/cookbook/pagination/ for reference.

tdstein:
Resolves #75
Related to #137


src/posit/connect/cursors.py
tdstein, line 1: This is a modified version of the implementation in `paginator.py`. A classic abstraction is possible, but it isn't necessary at this time.

If it's confusing having two separate modules with more-or-less the same exported classes, we could reorganize `curosrs.py` and `paginator.py` into a sub-module.

src/posit/connect/visits.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+            for result in results
+        )
+        return next(visits, None)
+
+
+def rename_params(params: dict) -> dict:
+    """Rename params from the internal to the external signature.
+
+    The API accepts `from` as a querystring parameter. Since `from` is a reserved word in Python, the SDK uses the name `start` instead. The querystring parameter `to` takes the same form for consistency.
```
tdstein, line 222: Note this change! Open to other ideas if there are strong opinions.

tests/posit/connect/__api__/v1/instrumentation/content/visits.json
```diff
@@ -0,0 +1,24 @@
+{
```
nealrichardson, line 1: We should have more than one page of mocks to test the feature, right?

src/posit/connect/cursors.py
```diff
@@ -0,0 +1,75 @@
...
+_MAX_PAGE_SIZE = 500
+
+
+@dataclass
+class Page:
+    paging: dict
+    results: List[dict]
+
+
+class Paginator:
```
nealrichardson, line 18: Is it confusing that we have two classes called Paginator, just in different submodules? Would it make sense to name these differently?

src/posit/connect/cursors.py
nealrichardson: I don't think we need to make them share code or inherit from a common thing. There are only 2.

PR #166, opened 2024-04-11
feat: adds usage

tdstein:
This is more or less a copy of the `visits` implementation. 

I am considering the best way to facade these resources behind a single interface. See https://github.com/posit-dev/posit-sdk-py/discussions/167

### DIFF
--- a/src/posit/connect/usage.py
+++ b/src/posit/connect/usage.py
@@ -1,0 +1,207 @@
+from __future__ import annotations
+
+from typing import List, overload
+
+from . import urls
+
+from .cursors import CursorPaginator
+from .resources import Resource, Resources
+
+
+class UsageEvent(Resource):
+    @property
+    def content_guid(self) -> str:
+        """The associated unique content identifier.
+
+        Returns
+        -------
+        str
+        """
+        return self["content_guid"]
+
+    @property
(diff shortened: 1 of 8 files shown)


src/posit/connect/usage.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        return self["data_version"]
+
+
+class Usage(Resources):
+    @overload
+    def find(
+        self,
+        content_guid: str = ...,
+        min_data_version: int = ...,
```
jonkeane, line 68: Do you have a test for something other than the default for `min_data_version`? Something that is a bit funny is that though the parameter is the same, for shiny the default (and max) is 1, for not-shiny content the default (and max) is 3

tdstein: These arguments are passed through as query parameters to the server, which performs all validation. If the server responds with a failure, then a ClientError is raised. 

```python
>>>  from posit import connect
>>> c = connect.Client()
>>> event = c.usage.find_one(min_data_version=-1)
Traceback (most recent call last):
...
posit.connect.errors.ClientError: The parameter min_data_version=-1 is less than the allowed minimum 0 (Error Code: 25, HTTP Status: 400 Bad Request)
```

PR #169, opened 2024-04-12
refactor: implements config using a singleton pattern

tdstein:
Since it is now clear that the config will not change after the initial setup, we can utilize a singleton pattern to reduce implementation complexity.


src/posit/connect/config.py
tdstein, line 1: The crux of the change is here. Everything else is a side effect of this.

tests/posit/connect/conftest.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from posit.connect import config
+
+
+def pytest_configure():
+    c = config.Config()
+    c.api_key = "12345"
+    c.url = "https://connect.example/__api__"
```
tdstein, line 9: This runs at the beginning of the test suite. Therefore, tests can assume these values when creating mocks or whatever else.

tests/posit/connect/test_config.py
```diff
@@ -1,49 +1,62 @@
...
+@pytest.fixture(autouse=True)
+def setup():
+    # capture the convert config settings
+    c = config.Config()
+    # reset the config before each test
+    config.reset()
+    # invoke the test
+    yield
+    # set the config back to previous state
+    config.Config.instance = c
```
tdstein, line 18: Since Config is a singleton, we must do a little more work to ensure a stable test state. This is only required for testing the Config directly.

tests/posit/connect/test_bundles.py
```diff
@@ -100,7 +97,7 @@ def test(self):
         )
 
         # setup
-        c = Client("12345", "https://connect.example")
+        c = Client()
```
tdstein, line 100: Now that Config is set up before tests run, we can forgo arguments here!

tests/posit/connect/test_config.py
```diff
@@ -1,49 +1,62 @@
...
+@pytest.fixture(autouse=True)
+def setup():
+    # capture the convert config settings
+    c = config.Config()
+    # reset the config before each test
+    config.Config.instance = None
+    # invoke the test
+    yield
+    # set the config back to previous state
+    config.Config.instance = c
```
tdstein, line 18: Since Config is a singleton, we must do a little more work to ensure a stable test state. This is only required for testing the Config directly.

nealrichardson:
While poking in `connectapi` this week, I ran across this function: https://github.com/rstudio/connectapi/blob/main/R/promote.R

It allows you to pull content from one Connect server and deploy it to another, as in a staging/production workflow. If `Config` is a singleton, will that be possible?

tdstein:
> While poking in `connectapi` this week, I ran across this function: https://github.com/rstudio/connectapi/blob/main/R/promote.R
> 
> It allows you to pull content from one Connect server and deploy it to another, as in a staging/production workflow. If `Config` is a singleton, will that be possible?

That's possible but requires a slightly different pattern to work as expected.


> While poking in `connectapi` this week, I ran across this function: https://github.com/rstudio/connectapi/blob/main/R/promote.R
> 
> It allows you to pull content from one Connect server and deploy it to another, as in a staging/production workflow. If `Config` is a singleton, will that be possible?

It's possible, but not using the same flow defined `promote.R`. That's probably a good reason not to do this, though.

PR #170, opened 2024-04-12
feat: adds metrics view events

### LINKED ISSUE #137: Content usage wrapper

tdstein:
The `metrics.views` package coalesces the existing instrumentation APIs into a single interface. The new event type `metrics.views.ViewEvent` encapsulates the `metrics.usage.UsageEvent` and `metrics.visits.VisitEvent`. The `ViewEvent` schema is a mashup of the two underlying schemas, with the exception of the `time` field on `VisitEvent`. In `ViewEvent` this field is removed and mapped to `started` and `ended`.  Otherwise, fields that do not exist in the underlying schema return `None`. 

Resolves #137

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -110,9 +110,7 @@ class Client:
         >>> from posit import (
         ...     connect,
         ... )
-        >>> client = (
-        ...     connect.Client()
-        ... )
+        >>> client = connect.Client()
         >>> client.metrics.views.find_one()
         {'content_guid': '2243770d-ace0-4782-87f9-fe2aeca14fc8', 'user_guid': '434f97ab-4b97-4443-8490-ed1052f37b29', 'variant_key': 'dZnFUBMR', 'rendering_id': '3029259', 'bundle_id': '49747', 'started': '2023-10-27T16:19:40Z', 'ended': '2023-10-27T16:19:40Z', 'data_version': 3, 'path': None}
         """
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -101,7 +101,12 @@ class Client:
     def metrics(self) -> metrics.Metrics:
         """The Metrics API interface.
 
-        The Metrics API is a software interface designed for capturing, retrieving, and managing quantitative data measurements. It is commonly used for monitoring and analyzing system performance, user behavior, and business processes. This API facilitates real-time data collection and accessibility, enabling organizations to make informed decisions based on key performance indicators (KPIs).
+        The Metrics API is designed for capturing, retrieving, and managing
+        quantitative measurements of Connect interactions. It is commonly used
+        for monitoring and analyzing system performance, user behavior, and
+        business processes. This API facilitates real-time data collection and
+        accessibility, enabling organizations to make informed decisions based
+        on key performance indicators (KPIs).
 
         Returns
         -------
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -5,15 +5,15 @@ from __future__ import annotations
 from requests import Response, Session
 from typing import Optional
 
-from . import hooks, me, urls
+from . import config, hooks, me, metrics, urls
 
 from .auth import Auth
 from .config import Config
 from .oauth import OAuthIntegration
 from .content import Content
-from .usage import Usage
+from .metrics.usage import Usage
 from .users import User, Users
-from .visits import Visits
+from .metrics.visits import Visits
 
 
 class Client:
@@ -98,12 +98,14 @@ class Client:
         return Content(config=self.config, session=self.session)
 
     @property
-    def usage(self) -> Usage:
-        return Usage(self.config, self.session)
+    def metrics(self) -> metrics.Metrics:
+        """The metrics API interface.
 
-    @property
-    def visits(self) -> Visits:
-        return Visits(self.config, self.session)
+        Returns
+        -------
+        metrics.Metrics
+        """
+        return metrics.Metrics(self.config, self.session)
 
     def __del__(self):
         """Close the session when the Client instance is deleted."""
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -104,6 +104,17 @@ class Client:
         Returns
         -------
         metrics.Metrics
+
+        Examples
+        --------
+        >>> from posit import (
+        ...     connect,
+        ... )
+        >>> client = (
+        ...     connect.Client()
+        ... )
+        >>> client.metrics.views.find_one()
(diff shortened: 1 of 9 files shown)


src/posit/connect/client.py
```diff
@@ -98,12 +98,30 @@ def content(self) -> Content:
...
 
-    @property
-    def visits(self) -> Visits:
-        return Visits(self.config, self.session)
+        The Metrics API is designed for capturing, retrieving, and managing
+        quantitative measurements of Connect interactions. It is commonly used
+        for monitoring and analyzing system performance, user behavior, and
+        business processes. This API facilitates real-time data collection and
+        accessibility, enabling organizations to make informed decisions based
+        on key performance indicators (KPIs).
```
nealrichardson, line 109: Is this true, or is this Copilot? ;)

tests/posit/connect/metrics/test_usage.py
```diff
@@ -69,15 +70,16 @@ def test(self):
         )
 
         # setup
-        c = Client("12345", "https://connect.example")
+        c = config.Config("12345", "https://connect.example")
+        session = requests.Session()
 
         # invoke
-        usage = c.usage.find()
+        events = usage.Usage(c, session).find()
```
nealrichardson, line 77: This doesn't seem right, right? I can't get here from `Client`?

nealrichardson (commented):
I'm struggling to read this, the tests aren't making clear to me how I (as a user) should use this and what I do with it.

src/posit/connect/client.py
tdstein: Haha. It isn't Co-Pilot. I crafted it, and I believe it to be true.

tests/posit/connect/metrics/test_usage.py
tdstein: I removed it in favor of the Views interface, but I'm unsure if that's the right call. What is your opinion? Should we continue exposing `Usage` and `Visits` even though `Views` is a superset of them?

nealrichardson:
Is this PR intended to solve https://github.com/posit-dev/posit-sdk-py/issues/137?

tdstein:
> Is this PR intended to solve #137?

Yes

tdstein:
> I'm struggling to read this, the tests aren't making clear to me how I (as a user) should use this and what I do with it.

The main interface is `client.metrics.views` which has fine `.find` and `.find_one`. The returned object is the `ViewEvent` which contains all of the information from the instrumentation APIs.

tests/posit/connect/metrics/test_usage.py
nealrichardson: I think we should wire them up all under `metrics`. Since they're in the API, it seems reasonable that we should provide a wrapper for them. Perhaps the user knows they only want one and not the other, and it's more efficient to just request the one they want. We prefer the one that just works and has everything, and we emphasize that in docs and guides.

As I was reading the API docs to try to understand the feature, I realized there's another place we probably want to expose this: on `ContentItem`, which would hit these APIs but filtered by guid. In that case, you won't want to query both endpoints because only one will work: one is for shiny (R only? IDK) and one is for everything else. That feels like a place where I don't want all three available, I don't want to have to think about which is right?

Does that make sense?

tdstein: It does. Based on my conversations with @kmasiello and other's it seems like understanding that aspect of the APIs is cumbersome. Aside from that, I really don't like the existing names. Maybe, namespace them under `metrics.instrumentation` (e.g., `metrics.instrumentation.usage` and `metrics.instrumentation.visits`).

Adding the hook in `ContentItem` is a great idea.

nealrichardson: > Aside from that, I really don't like the existing names. 

Yeah, agreed. Honestly I can't tell the difference between `Views`, `Visits`, and `Usage`. 

> Maybe, namespace them under `metrics.instrumentation` (e.g., `metrics.instrumentation.usage` and `metrics.instrumentation.visits`).

I guess you could 🤷 . So `metrics.instrumentation` would yield another object that has `usage` and `visits` attributes?

Do we need Views under `metrics`? Do we have other metrics? I guess this is top-level, so we could have other things like system metrics?

Would it make sense to put Views off of Content, even though that's not where they're mounted in the API? Like `client.content.views.find()`? They are views of content, yeah?

tdstein: > Yeah, agreed. Honestly I can't tell the difference between Views, Visits, and Usage.

We should use this as an opportunity to improve the API. If the `Views` abstraction is well received, we should introduce that in a future version of Connect and drop the rest of the instrumentation API.

> Do we have other metrics? 

Nothing else at the moment, but we should. @christierney / @Lytol - what else are we pushing into Prometheus? It seems right to me to separate the namespace for metrics to declutter the non-metrics part of the API.

>  Would it make sense to put Views off of Content.

We can. It's the existing implementation plus a filter on content_guid.

nealrichardson: > > Yeah, agreed. Honestly I can't tell the difference between Views, Visits, and Usage.
>
> We should use this as an opportunity to improve the API. If the Views abstraction is well received, we should introduce that in a future version of Connect and drop the rest of the instrumentation API.

Agreed. I just meant more about naming, I couldn't tell which was the one that contained the others since they all sounds the same. But if we're only presenting one name, I think it's fine.

> > Would it make sense to put Views off of Content.
> 
> We can. It's the existing implementation plus a filter on content_guid.

Let me clarify. I think there are two different ideas:

1. Add `content.get(guid).views.find()`, which is equivalent to the current `metrics.views.find(content_guid = guid)`. That's from a previous comment. 
2. What I meant here was: instead of `client.metrics.views`, should it be exposed via `client.content.views` and skip the `metrics` path entirely? If `views` are always about Content, maybe it makes logical sense to access them via content. (Or if we have other metrics to expose, we could do both?)

tdstein: Add #181

PR #179, opened 2024-04-23
feat: adds bundle create and download

### LINKED ISSUE #108: Bundles.create() by upload
https://docs.posit.co/connect/api/#post-/v1/content/-guid-/bundles

Let's defer dealing with manifests and all that. Just focus on taking a tarball and uploading it.

tdstein:
Resolves #108


src/posit/connect/bundles.py
```diff
@@ -99,14 +96,66 @@ def delete(self) -> None:
         url = urls.append(self.config.url, path)
         self.session.delete(url)
 
+    def download(self) -> bytes:
+        """Download a bundle.
+
+        Returns
+        -------
+        bytes
+            Archive contents in bytes representation.
```
nealrichardson, line 105: Will there ever be a use for this? Won't you always want to write this to a file?

src/posit/connect/bundles.py
```diff
@@ -1,17 +1,14 @@
 from __future__ import annotations
 
-from typing import List
+from typing import BinaryIO, List
 
 from requests.sessions import Session as Session
```
nealrichardson, line 5: `import Session as Session`?

src/posit/connect/bundles.py
```diff
@@ -99,14 +96,66 @@ def delete(self) -> None:
...
 
+    def create(self, data: BinaryIO | bytes) -> Bundle:
+        """Create a bundle.
+
+        Create a bundle by upload via archive format.
+
+        Parameters
+        ----------
+        data : BinaryIO | bytes
+            Archive contents in BinaryIO or bytes representation.
```
nealrichardson, line 134: Similar to above, would it make sense be able to just pass a file path? I would imagine that `requests` would have some handling for that kind of form posting too.

src/posit/connect/bundles.py
tdstein: We can inspect in memory, but I'm not sure if that will be useful. @mmarchetti, what are your thoughts? Do we need to provide support for inspecting the tarball in memory?

src/posit/connect/bundles.py
tdstein: Ha, that was vscode auto import. I'll fix.

src/posit/connect/bundles.py
tdstein: Yes, looks like requests supports providing a filename as a string. I'll try that out.

src/posit/connect/bundles.py
mmarchetti: I think a download to file option is probably sufficient. If we do add an in-memory option, it would be better to return a BytesIO rather than to read the contents immediately.

src/posit/connect/bundles.py
mmarchetti: There is precedent in a number of libraries for functions that accept a file handle or filename.

tdstein: Do you know any off the top of your head?

mmarchetti: https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html is an example

nealrichardson (approved):
LGTM, seems straightforward

PR #183, opened 2024-04-25
feat: adds bundle/content deploy and task management

### LINKED ISSUE #110: Deploy a bundle
https://docs.posit.co/connect/api/#post-/v1/content/-guid-/deploy

### LINKED ISSUE #111: Task polling
https://docs.posit.co/connect/api/#get-/v1/tasks/-id-

### LINKED ISSUE #112: Create content (epic)
Following https://docs.posit.co/connect/cookbook/deploying/, we need a few things:

- [x] #106 
- [x] #109 
- [x] #108 
- [x] #110 
- [x] #111

tdstein:
Resolves #110 
Resolves #111 
Resolves #112

### DIFF
--- a/src/posit/connect/bundles.py
+++ b/src/posit/connect/bundles.py
@@ -5,7 +5,7 @@ import requests
 
 from typing import BinaryIO, List
 
-from . import config, resources, urls
+from . import config, resources, tasks, urls
 
 
 class BundleMetadata(resources.Resource):
@@ -95,6 +95,21 @@ class Bundle(resources.Resource):
         path = f"v1/content/{self.content_guid}/bundles/{self.id}"
         url = urls.append(self.config.url, path)
         self.session.delete(url)
+
+    def deploy(self) -> tasks.Task:
+        """Deploy the bundle.
+
+        Returns
+        -------
+        tasks.Task
+            The task for the deployment.
+        """
+        path = f"v1/content/{self.content_guid}/deploy"
+        url = urls.append(self.config.url, path)
+        response = self.session.post(url, json={"bundle_id": self.id})
+        result = response.json()
+        ts = tasks.Tasks(self.config, self.session)
+        return ts.get(result["task_id"])
 
     def download(self, output: io.BufferedWriter | str):
         """Download a bundle.
--- a/src/posit/connect/bundles.py
+++ b/src/posit/connect/bundles.py
@@ -3,7 +3,7 @@ import io
 
 import requests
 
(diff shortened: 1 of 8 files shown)


src/posit/connect/bundles.py
```diff
@@ -3,9 +3,9 @@
 
 import requests
 
-from typing import List
+from typing import BinaryIO, List
```
nealrichardson, line 6: I don't see this used in the diff, is it needed?

```suggestion
from typing import List
```

src/posit/connect/tasks.py
```diff
@@ -0,0 +1,204 @@
...
+        sleep : int, optional
+            Maximum number of seconds to wait between status checks.
+
+        Examples
+        --------
+        >>> task.wait_for()
+        None
+
+        Wait five seconds between status checks.
+        >>> task.wait_for(5)
```
nealrichardson, line 145: When I read this, I thought this would wait for 5 seconds and then return. But it turns out that it will wait as long as it takes to finish, just updating at 5-second intervals.

src/posit/connect/tasks.py
```diff
@@ -0,0 +1,204 @@
...
+        None
+        """
+        while not self.is_finished:
+            self.update(wait=sleep)
```
nealrichardson, line 149: I know the API supports this, but it's odd to me that we'd push the `sleep` to the server. I.e. with `wait_for(5)`, each request is going to take 5 seconds to respond. Seems odd to consume server resources for that waiting. Why not do the sleeping on the client side and always send wait=0 (which is the default anyway)?

The other thing you might do is pass the `first` argument, set to `len(self.output)` so that you're only returning newly generated lines of log output. You'd have to do something different than just `super().update(**result)` with that, have to be sure to append output and update the rest. But I could see that for a large task (IDK, lots of package installation output), not fetching all lines every time could be important.

tdstein: I had the same reaction when developing this. The one advantage of the current implementation is that the API will return once the task is finished. So, if you call with `wait=60`, the method will return in less than 60 seconds if the task finishes. 

But I also agree. It seems like an anti-pattern to consume server threads waiting for the task to complete.

tdstein: Added #184

src/posit/connect/tasks.py
tdstein: It does read that way. I'm going to remove the argument for now an address it in #184

Issue #165, opened 2024-04-10
ContentItem.owner


nealrichardson:
Similar to #164 but less complicated because `GET v1/content?include=owner` always includes the owner.

If `owner` is not in the item's data, the owner `@property` should `GET v1/users/{owner_guid}`, insert the response into the data, and return it.

brooklynbagel:
Is this for the `find` method? Maybe this a separate issue, but I noticed the type given for the `owner` property mismatches with what Connect provides:

```
owner 🆁: {
  Basic details about the owner of this content item. Included in responses when include=owner is provided.

  guid: uuid
  The user's GUID, or unique identifier, in UUID [RFC4122](https://www.rfc-editor.org/rfc/rfc4122) format

  username: string
  The user's username

  first_name: string
The user's first name

  last_name: string
  The user's last name example: 'Doe'
}
```

https://docs.posit.co/connect/api/#get-/v1/content

versus:

https://github.com/posit-dev/posit-sdk-py/blob/62ff3596c56217d3f5fb2d57d95bde53f5655b4f/src/posit/connect/content.py#L193-L195

nealrichardson:
Yeah good catch, I think you're right.

brooklynbagel:
I can submit a PR, should it be linked to this issue or elsewhere?

brooklynbagel:
I think it's a simple fix, I can open a PR and either link to this issue or create a new one.

nealrichardson:
That would be great, thanks!

PR #190, opened 2024-05-03
feat: add dataclass to represent ContentItem owner

brooklynbagel:
Addresses the owner property of ContentItem as discussed in #165 

I ran all the steps required plus a manual test with the following output:

```python

with Client() as client:
  all_content = client.content.find()
  print(all_content[0].owner)
# {'guid': '867f7ef5-6cb4-4a4d-908d-21f7ccacf273', 'username': 'michael.beigelmacher@posit.co', 'first_name': 'Michael', 'last_name': 'Beigelmacher'}
```

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -15,6 +15,26 @@ from .permissions import Permissions
 from .resources import Resources, Resource
 
 
+class ContentItemOwner(Resource):
+    """The owner of a piece of content."""
+
+    @property
+    def guid(self) -> str:
+        return self.get("guid")  # type: ignore
+
+    @property
+    def username(self) -> str:
+        return self.get("username")  # type: ignore
+
+    @property
+    def first_name(self) -> Optional[str]:
+        return self.get("first_name")  # type: ignore
+
+    @property
+    def last_name(self) -> Optional[str]:
+        return self.get("last_name")  # type: ignore
+
+
 class ContentItem(Resource):
     """A piece of content."""
 
@@ -191,7 +211,7 @@ class ContentItem(Resource):
         return self.get("owner_guid")  # type: ignore
 
     @property
-    def owner(self) -> str:
+    def owner(self) -> ContentItemOwner:
         return self.get("owner", {})  # type: ignore
 
     @property


tdstein (changes requested):
Thanks for the pull request @brooklynbagel. We have taken a different approach to modeling schemas in this project. Instead of using a `@dataclass` please modify the implementation to inherit from `resources.Resource` and create `@property` attributes for each field. You can mimic the `ContentItem`.

If you are interested in learning more about this approach, there are various discussions in the issue history. But, the basic idea is that this pattern provides an escape hatch for forward and backwards compatibility between server (Connect) versions. It also provides an intuitive experience when converting collections of these datatypes to Pandas and Polars data frames.

brooklynbagel:
> If you are interested in learning more about this approach, there are various discussions in the issue history. But, the basic idea is that this pattern provides an escape hatch for forward and backwards compatibility between server (Connect) versions. It also provides an intuitive experience when converting collections of these datatypes to Pandas and Polars data frames.

Interesting, this is something I'll look. Thanks for the note and I wasn't sure because there were a couple of places where `@dataclass` are used. See 
https://github.com/posit-dev/posit-sdk-py/blob/19ec089f9a2af60363f1f3b7400c75cace1f49a2/src/posit/connect/paginator.py#L11-L25 and 
https://github.com/posit-dev/posit-sdk-py/blob/19ec089f9a2af60363f1f3b7400c75cace1f49a2/src/posit/connect/cursors.py#L12-L15

brooklynbagel:
I can mark this as ready for review but I'm not sure why CI is failing for Python 3.12 and if it's an error we can ignore.

```
Run orgoro/coverage@v3.1
  with:
    coverageFile: coverage.xml
    thresholdAll: 0.8
    token: ***
    thresholdNew: 0
    thresholdModified: 0
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.12.3/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.12.3/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.3/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.3/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.3/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.12.3/x64/lib
comparing commits: base 62ff3596c56217d3f5fb2d57d95bde53f5655b4f <> head 8e2e479cfd64d350739c84eade90abe87bf43831
git new files: [] modified files: ["src/posit/connect/content.py"]
source: /home/runner/work/posit-sdk-py/posit-sdk-py
Results
  Average coverage 96% ✅
  No covered new files in this PR 
  Modified files coverage ✅
HttpError: Resource not accessible by integration
    at /home/runner/work/_actions/orgoro/coverage/v3.1/webpack:/typescript-action/node_modules/@octokit/request/dist-node/index.js:86:1
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

tdstein:
> Thanks for the note and I wasn't sure because there were a couple of places where `@dataclass` are used.

Ah, since those are internal, I used dataclasses. However, on second thought, it would be better to use a resource type since the API response schema can change between Connect versions.

Thanks for calling those out!

tdstein (approved):

tdstein:
> I can mark this as ready for review but I'm not sure why CI is failing for Python 3.12 and if it's an error we can ignore.

Yes, we can ignore that. We have an issue to resolve that. It's a limitation of forks versus branches in GitHub Actions.

brooklynbagel:
> > I can mark this as ready for review but I'm not sure why CI is failing for Python 3.12 and if it's an error we can ignore.
> 
> Yes, we can ignore that. We have an issue to resolve that. It's a limitation of forks versus branches in GitHub Actions.

Got it, I'll keep that in mind if I submit any future PRs. I don't have write access so I cannot perform the merge myself.

PR #195, opened 2024-05-24
feat: Check for X-Deprecated-Endpoint header

nealrichardson:
Analogous to https://github.com/rstudio/connectapi/pull/268. `posit-sdk` shouldn't be calling deprecated endpoints--when we deprecate endpoints that it uses, we'll make sure to check the Connect server version and call the correct endpoints accordingly, so that it only calls deprecated endpoints on older Connect versions where they aren't deprecated. The only circumstance where `posit-sdk` would hit a deprecated endpoint would be if you're pointing at a newer version of Connect but an old version of`posit-sdk` that doesn't know about the new, non-deprecated endpoints. So, when receiving a X-Deprecated-Endpoint response header, the recommendation to the user is to upgrade `posit-sdk`.

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -5,15 +5,13 @@ from __future__ import annotations
 from requests import Response, Session
 from typing import Optional
 
-from . import config, hooks, me, metrics, tasks, urls
+from . import hooks, me, metrics, tasks, urls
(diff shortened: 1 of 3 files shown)


tdstein (approved):
Nice!

PR #199, opened 2024-05-30
ci: conslidate duplicate workflows

tdstein:


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,25 @@
+name: Coverage
+on:
+  - pull_request
```
nealrichardson, line 3: We don't want coverage running on `main`, only on PRs?

tdstein: Coverage report is only valuable on pull requests.

src/posit/connect/hooks.py
```diff
@@ -14,14 +14,18 @@ def handle_errors(response: Response, *args, **kwargs) -> Response:
             message = data["error"]
             http_status = response.status_code
             http_status_message = responses[http_status]
-            raise ClientError(error_code, message, http_status, http_status_message)
+            raise ClientError(
+                error_code, message, http_status, http_status_message
+            )
```
tdstein, line 19: This is all auto-formatting from Ruff

PR #210, opened 2024-06-05
build: adds integration tests tooling

tdstein:
## Overview

Adds integration test infrastructure. The architecture is designed to provide fast execution across the Connect version grid within a developer environment. On my machine, the full test grid is completed in 30 to 45 seconds.

## Quick start

Run `make help` from the `integration` directory.

Run `make latest` to run the test suite against 2024.05.0.

## Architecture summary

A new Makefile is defined, which provides rules for executing the integration test suite against Connect. The Connect version is parameterized across `Makefile` and `docker-compose.yaml` via the `CONNECT_VERSION` environment variable. The Connect environment is configured for "bootstrapping" via `CONNECT_BOOTSTRAP_ENABLED`. The existing `rsconnect` implementation obtains the `CONNECT_API_KEY ` in `make test`. Ideally, this will be replaced in the future with an SDK implementation.

- `Makefile`: defines commands for integration test invocation.
- `docker-compose.yaml`: defines environment integration testing and invokes test suite
- `.github/workflows/ci.yaml`: adds integration-test job that runs test suite in a matrix.


src/posit/connect/client.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import requests
```
nealrichardson, line 5: What is this for?

.github/workflows/ci.yaml
```diff
@@ -39,6 +39,36 @@ jobs:
...
+          - 2023.06.0
+          - 2023.05.0
+          - 2023.01.1
+          - 2023.01.0
+          - 2022.12.0
+          - 2022.11.0
+    steps:
+    - uses: actions/checkout@v4
+    - uses: docker/setup-buildx-action@v3
+    - uses: docker/setup-qemu-action@v3
```
nealrichardson, line 66: What do we need qemu for? Everything in this stack is x86 right?

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3
```
nealrichardson, line 1: What is this Dockerfile used for?

.github/workflows/ci.yaml
```diff
@@ -39,6 +39,36 @@ jobs:
...
+          - 2023.12.0
+          - 2023.10.0
+          - 2023.09.0
+          - 2023.07.0
+          - 2023.06.0
+          - 2023.05.0
+          - 2023.01.1
+          - 2023.01.0
+          - 2022.12.0
+          - 2022.11.0
```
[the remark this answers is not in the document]
tdstein: They all complete around the same time when run concurrently. I think it should be whatever our support window is.

integration/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+    # Run integration test suite.
+    #
+    # Target is relative to the ./integration directory, not the project root
+    # directory. The execution base directory is determined by the 'WORKDIR'
+    # in the Dockerfile.
+    command: make -C ./integration test
+    environment:
+        - CONNECT_BOOTSTRAP_SECRETKEY=${CONNECT_BOOTSTRAP_SECRETKEY}
+        # Port 3939 is the default port for Connect
+        - CONNECT_SERVER=http://connect:3939
```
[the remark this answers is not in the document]
tdstein: The hostname is the relative service name. So https://github.com/posit-dev/posit-sdk-py/blob/f8e38a79a7a33cae99f56c3431fe90170a281c2b/integration/docker-compose.yaml#L22

.github/workflows/ci.yaml
tdstein: Oh, good catch. I'll try removing it.

Dockerfile
tdstein: This is the test runner for the "tests" service in the compose file. https://github.com/posit-dev/posit-sdk-py/blob/f8e38a79a7a33cae99f56c3431fe90170a281c2b/integration/docker-compose.yaml#L3

PR #211, opened 2024-06-06
ci: integration test reports

tdstein:
See https://github.com/posit-dev/posit-sdk-py/runs/25953636649


tdstein:
Creating a new one to deal with Netlify weirdness

PR #216, opened 2024-06-10
ci: add production site deployment

tdstein:


nealrichardson:
What is the resulting production URL? I'm guessing we want this under docs.posit.co, maybe even `https://docs.posit.co/posit-sdk-py`. That isn't netlify is it?

tdstein:
> What is the resulting production URL? I'm guessing we want this under docs.posit.co, maybe even `https://docs.posit.co/posit-sdk-py`. That isn't netlify is it?

It writes to the production Netlify site. Which ops will route to the appropriate docs url.

annikakrilov (approved):

PR #223, opened 2024-06-12
docs: set development status to alpha

tdstein:
This is arbitrary, but I believe we have moved a pre-alpha state to alpha an alpha state. This information will be visible in  on PyPI on the next release.


nealrichardson (approved):
Sure. I don't think this means anything.

PR #224, opened 2024-06-12
feat: always resolve content owner

### LINKED ISSUE #165: ContentItem.owner

tdstein:
Resolves #165


tests/posit/connect/__api__/v1/content/f2f37341-e21d-3d80-c698-a935ad614066.json
```diff
@@ -37,7 +37,13 @@
     "default_py_environment_management": null,
     "run_as": null,
     "run_as_current_user": false,
-    "owner_guid": "87c12c08-11cd-4de1-8da3-12a7579c4998",
+    "owner_guid": "20a79ce3-6e87-4522-9faf-be24228800a4",
+    "owner": {
```
nealrichardson, line 41: I would back this change out. The API wouldn't return `owner` in this request.

src/posit/connect/content.py
```diff
@@ -146,6 +147,14 @@ def bundles(self) -> Bundles:
     def permissions(self) -> Permissions:
         return Permissions(self.config, self.session, self.guid)
 
+    @property
+    def owner(self) -> ContentItemOwner:
+        if "owner" in self:
+            return self["owner"]
+        user = Users(self.config, self.session).get(self.owner_guid)
+        owner = ContentItemOwner(self.config, self.session, **user)
+        return owner
```
nealrichardson, line 156: How about this? We should cache the owner User.

```suggestion
        if "owner" not in self:
            # It is possible to get a content item that does not contain owner.
            # "owner" is an optional additional request param.
            # If it's not included, we can retrieve the information by `owner_guid`
            self["owner"] = Users(self.config, self.session).get(self.owner_guid)
        return ContentItemOwner(self.config, self.session, **self["owner"])
```

tdstein: Nice. 👏🏻

tests/posit/connect/test_content.py
```diff
@@ -165,6 +187,30 @@ def test_tags(self):
...
+            f"https://connect.example/__api__/v1/users/20a79ce3-6e87-4522-9faf-be24228800a4",
+            json=load_mock(
+                f"v1/users/20a79ce3-6e87-4522-9faf-be24228800a4.json"
+            ),
+        )
+
+        c = Client("12345", "https://connect.example")
+        item = c.content.get("f2f37341-e21d-3d80-c698-a935ad614066")
+        owner = item.owner
+        assert owner.guid == "20a79ce3-6e87-4522-9faf-be24228800a4"
```
nealrichardson, line 211: Other things to assert: 

* owner is ContentItemOwner (if we care about that)
* The user GET on L201 was called once
* Do item.owner again and confirm that L201 has still only been called once.

nealrichardson (approved):
One last suggestion on the test but otherwise LGTM!

PR #226, opened 2024-06-17
feat: get content for user

tdstein:
Adds scoped content calls off of the User resource (e.g., `client.users.find_one().content.find_one()`.


integration/tests/posit/connect/test_content.py
```diff
@@ -10,6 +11,11 @@ def setup_class(cls):
             access_type="acl",
         )
 
+    @classmethod
+    def teardown_class(cls):
+        cls.item.delete()
+        assert cls.client.content.count() == 0
```
toph-allen, line 17: I'm new to the architecture of this package — for my own learning, what does this do?

src/posit/connect/users.py
```diff
@@ -17,6 +18,8 @@ class User(Resource):
 
     Attributes
     ----------
+    content: Content
+        Content resource for user. Sets the content item owner identifier (`owner_guid`) to `self.guid`.
```
toph-allen, line 22: So this lets you interact with content that is owned by the user?

toph-allen: Seemed almost too obvious to ask, but I'm asking to distinguish, does it include stuff they collaborate on?

toph-allen (approved):
Just a few comments, just questions, mostly for my own comprehension.

integration/tests/posit/connect/test_content.py
tdstein: `cls.item` is the content item created in the `setup_class` method. L16 performs a delete action, and L17 asserts that the total number of content items on the server is 0. The user used for testing has administrator privileges.

The general idea here is to try and ensure a "clean" server state between test classes. That way, other test classes would not need to be aware of each other. I'm not sure this is viable in the long term, but it's okay for now.

src/posit/connect/users.py
tdstein: Eh, this isn't a good description. Is this better?

```suggestion
        A content resource scoped to this user.
```


Check this out for context: https://github.com/posit-dev/posit-sdk-py/blob/58a8b9f6ced7cd0c3fd8c8259692fcf33b4b900b/src/posit/connect/content.py#L633-L646

The requests to GET v1/content are scoped with the query param `owner_guid={the guid of this user}`. If my understanding is correct, this will only include content they "own". Not those that they collaborate on.

PR #227, opened 2024-06-24
feat: adds groups

tdstein:
Unit tests are sparse since we have integration tests now.

### DIFF
--- a/integration/tests/posit/connect/test_groups.py
+++ b/integration/tests/posit/connect/test_groups.py
@@ -1,0 +1,23 @@
+from posit import connect
+
+
+class TestGroups:
+    def setup_class(cls):
+        cls.client = connect.Client()
+        cls.item = cls.client.groups.create(name="Friends")
+
+    def teardown_class(cls):
+        cls.item.delete()
+        assert cls.client.groups.count() == 0
+
+    def test_count(self):
+        assert self.client.groups.count() == 1
+
+    def test_get(self):
+        assert self.client.groups.get(self.item.guid)
+
+    def test_find(self):
+        assert self.client.groups.find() == [self.item]
+
+    def test_find_one(self):
+        assert self.client.groups.find_one() == self.item
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -14,6 +14,7 @@ from .content import Content
 from .metrics import Metrics
 from .tasks import Tasks
 from .users import User, Users
+from .groups import Groups
 
 
 class Client:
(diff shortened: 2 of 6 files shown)


src/posit/connect/groups.py
```diff
@@ -0,0 +1,201 @@
...
+        unique_id: str | None
+
+        Returns
+        -------
+        Group
+        """
+        ...
+        body = dict(*args, **kwargs)
+        path = "v1/groups"
+        url = urls.append(self.config.url, path)
```
jonkeane, line 93: Any reason to have `path` as a separate variable?

tdstein: not here, but comes into play downstream: https://github.com/posit-dev/posit-sdk-py/pull/228/files#diff-b7d2a2eb92d17a5b4aff237de4af47283494386561a1e96fea70f0d5ba4201d8R207-R211

jonkeane (approved):
This is looking good, one minor comment. (Sorry I didn't use the actual review flow with that one, wasn't sure how long I would be able to pay attention to this and didn't want comments dangling)

> Unit tests are sparse since we have integration tests now.

The test you have there is probably good enough, but IMHO and generally: leaning too heavily on integration tests will make development much harder. Having robust unit tests means it's quick + easy to get started and poke at it without relying on throwing it at CI and waiting for that to churn.

integration/tests/posit/connect/test_groups.py
```diff
@@ -0,0 +1,23 @@
...
+        assert self.client.groups.count() == 1
+
+    def test_get(self):
+        assert self.client.groups.get(self.item.guid)
+
+    def test_find(self):
+        assert self.client.groups.find() == [self.item]
+
+    def test_find_one(self):
+        assert self.client.groups.find_one() == self.item
```
toph-allen, line 23: So concise 😍

src/posit/connect/client.py
```diff
@@ -94,6 +95,17 @@ def oauth(self) -> OAuthIntegration:
         """
         return OAuthIntegration(config=self.config, session=self.session)
 
+    @property
+    def groups(self) -> Groups:
+        """The groups resource interface.
```
toph-allen, line 100: My gut response is that I'd want to see more descriptive text here, if this is used as doc text in the CLI. I don't know if "groups resource interface" will be a particularly meaningful term for our users (it isn't for me).

I see that some of this (`resource`) is terminology used throughout the package, and maybe it'll make more sense to me as I get to know it.

toph-allen (approved):
Looks good. I haven't looked at the Python SDK code much yet, and it's nice & very concise. :)

src/posit/connect/client.py
tdstein: Good call out. The entire Client class could use some documentation love.

PR #233, opened 2024-07-16
feat: content environment variables

tdstein:
WIP - Review not required. Just showing how things are going.


src/posit/connect/content.py
```diff
@@ -369,6 +375,11 @@ def deploy(self) -> tasks.Task:
         ts = tasks.Tasks(self.config, self.session)
         return ts.get(result["task_id"])
 
+    def restart(self) -> None:
+        random_hash = secrets.token_hex(32)
+        self.environment_variables[random_hash] = random_hash
+        del self.environment_variables[random_hash]
```
tdstein, line 381: Here's the basic concept. This will branch based on content type.

toph-allen: This came up with Brian on the call, but does calling these things without waiting for Connect to give some kind of response ever cause any issues like orphaned variables?

toph-allen: Perhaps better to prepend the random hash with `.CONNECT_RESTART_TMP_` so that, in some edge case where it does get orphaned, it's clear what it is.

tdstein: It's a synchronous call, so the next line does execute until the server accepts the response and returns a 2xx status code. I haven't observed any issues yet with testing, but adding the prefix is a good idea to provide some self-documentation of what's happening.

PR #236, opened 2024-07-22
feat: content render and restart

tdstein:
Add support for `content.restart()` and `content.refresh()`.


integration/resources/bundles/example-flask-minimal/hello.py
tdstein, line 1: The integration/resources folder contains bundles used for testing.

pyproject.toml
```diff
@@ -25,6 +25,9 @@ dependencies = ["requests>=2.31.0,<3"]
 Source = "https://github.com/posit-dev/posit-sdk-py"
 Issues = "https://github.com/posit-dev/posit-sdk-py/issues"
 
+[tool.mypy]
+exclude = "integration/resources/*"
```
tdstein, line 29: Ignore the resources folder with static type checking.

pyproject.toml
```diff
@@ -34,6 +37,7 @@ version_file = "src/posit/_version.py"
 
 [tool.ruff]
 line-length = 79
+exclude = ["integration/resources/*"]
```
tdstein, line 40: Ignore the resources folder when linting

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
         Tags associated with the content item.
     """
 
+    # CRUD Methods
+
+    def delete(self) -> None:
+        """Delete the content item."""
+        path = f"v1/content/{self.guid}"
+        url = urls.append(self.config.url, path)
+        self.session.delete(url)
```
tdstein, line 149: This is an unfortunate git diff resolution. This method already exists.

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
...
+
+        Notes
+        -----
+        This method is identical to `refresh` and exists to provide contextual clarity. Both methods produce identical results. When working with applications, natural language prefers "restart this content" instead of "refresh this content" since applications require a system process. When writing software that operates on multiple types of content (e.g., applications, documents, scripts, etc.), you may use either 'restart' or 'refresh' to achieve the same result.
+
+        Examples
+        --------
+        >>> restart()
+        """
+        return self._re_whatever()
```
tdstein, line 210: These methods are new. The rest of the code block highlighted in the Git diff already exists and is reordered. Unfortunately, Git isn't presenting it that way.

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
...
+            self.environment_variables.delete(key)
+            # GET via the base Connect URL to force create a new worker thread.
+            url = urls.append(dirname(self.config.url), f"content/{self.guid}")
+            self.session.get(url)
+            return None
+
+        warnings.warn(
+            f"Content '{self.guid}' with application mode '{self.app_mode}' does not require restarts"
+        )
+        return None
```
tdstein, line 356: This is also new.

src/posit/connect/content.py
```diff
@@ -160,6 +382,10 @@ def owner(self) -> ContentItemOwner:
             )
         return ContentItemOwner(self.config, self.session, **self["owner"])
 
+    @property
+    def _variants(self) -> Variants:
+        return Variants(self.config, self.session, self.guid)
```
tdstein, line 387: This is intentionally marked as private. We don't want customers to interact directly with the variants API since it isn't a v1 API.

src/posit/connect/env.py
```diff
@@ -0,0 +1,220 @@
...
+        ----------
+        key : str
+            The name of the environment variable to delete.
+
+        Examples
+        --------
+        >>> vars = EnvVars(config, session, content_guid)
+        >>> del vars["DATABASE_URL"]
+        """
+        self.update({key: None})
```
tdstein, line 61: These methods are included to provide an idiomatic interface for interfacing with environments. Unfortunately, the __getitem__ dunder method cannot be implemented since the environments API doesn't return values.

src/posit/connect/env.py
```diff
@@ -0,0 +1,220 @@
...
+                    raise TypeError(
+                        f"update expected a {MutableMapping} or {Iterable}, got {type(other)}"
+                    )
+            else:
+                raise TypeError(
+                    f"update expected a {MutableMapping} or {Iterable}, got {type(other)}"
+                )
+
+        if kwargs:
+            d.update(kwargs)
```
tdstein, line 215: This logic mimics the dict.update behavior.

integration/tests/posit/connect/test_content.py
```diff
@@ -37,3 +40,48 @@ def test_content_item_owner_from_include(self):
...
+        # deploy bundle
+        task = bundle.deploy()
+        task.wait_for()
+        # refresh content
+        task = content.refresh()
+        if task:
+            task.wait_for()
+
+        # delete content
+        content.delete()
```
tdstein, line 87: I'm still working on adding support for the skipped versions of Connect, but I'll make a separate change. Theres a bit involved to make it work, so it deserves its own pull request.

src/posit/connect/content.py
tdstein, line 356: @mmarchetti - I am adding you for this section in particular. This is the crux of the change, and I want to check if the business logic is reasonable.  The big caveat to this approach is upgrading the app_mode checks over time to accommodate new application modes. I've been tinkering with adding information to the warning message that directs users to open a GitHub issue as a lazy way to know if this happens.

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
...
+        None
+        """
+        path = f"v1/content/{self.guid}/deploy"
+        url = urls.append(self.config.url, path)
+        response = self.session.post(url, json={"bundle_id": None})
+        result = response.json()
+        ts = tasks.Tasks(self.config, self.session)
+        return ts.get(result["task_id"])
+
+    def refresh(self) -> Task | None:
```
toph-allen, line 174: I still feel like `refresh` and `restart` don't quite map well onto the problem space for me, but also like I haven't found the right way to explain how I'm thinking about it. Gonna take this opportunity to take another run at it, but don't feel like this is a comment on the PR itself.

We wanted `refresh` to be a method that's a good catch-all term that works generically for all content types.

I think we included `restart` because we reasoned that people with interactive content types might look for that verb.

I think similarly R Markdown and Quarto users will look for a `render` verb — that's definitely the preeminent term in that world (see the second heading of the Quarto [Getting Started](https://quarto.org/docs/get-started/hello/rstudio.html#rendering) guide).

I'm fine using `refresh` as a catch-all term, but I think it feels to me like we should _either_ have both `restart` and `render`, or have neither. I think it's just having one of them which feels off to me. 🤔

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
...
+
+        Submit a refresh request to the server for the content. After submission, the server executes an asynchronous process to refresh the content. This is useful when content is dependent on external information, such as a dataset.
+
+        See Also
+        --------
+        restart
+
+        Notes
+        -----
+        This method is identical to `restart` and exists to provide contextual clarity. Both methods produce identical results. When working with documents, natural language prefers "refresh this content" instead of "restart this content" since documents do not require a system process. When writing software that operates on multiple types of content (e.g., applications, documents, scripts, etc.), you may use either 'refresh' or 'restart' to achive the same result.
```
toph-allen, line 185: I like this way of discussing it.

src/posit/connect/content.py
toph-allen: LOL `._re_whatever()`.

integration/resources/bundles/example-flask-minimal/hello.py
nealrichardson: Should there be `connect` somewhere in the directory tree?

integration/tests/posit/connect/test_env.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+
+    @classmethod
+    def teardown_class(cls):
+        cls.content.delete()
+        assert cls.client.content.count() == 0
+
+    def test_clear(self):
+        self.content.environment_variables.create("KEY", "value")
+        assert self.content.environment_variables.find() == ["KEY"]
```
nealrichardson, line 21: Did you consider a dict-like interface for env vars? IDK if it would make sense, but it would look more like

```
        self.content.environment_variables["KEY"] = "value"
        assert self.content.environment_variables.keys() == ["KEY"]
```

which feels natural to type.

Maybe that could be sugar on top of the CRUD methods you've wrapped here? Maybe it's a bad idea? 🤷

src/posit/connect/content.py
toph-allen: In the R implementation, the already-implemented `render` app modes return an existing task object, so it's easy to return it. Do you think this is a reasonable point to diverge?

In R, I had moved towards just having the `render` function return a task object, but I'm not sure if consistency is better, particularly because interactive and rendered content are not different types.

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
...
+
+        Sends a restart request to the server for the content. Once submitted, the server performs an asynchronous process to restart the content. This is particularly useful when the content relies on external information loaded into application memory, such as datasets. Additionally, restarting can help clear memory leaks or reduce excessive memory usage that might build up over time.
+
+        See Also
+        --------
+        refresh
+
+        Notes
+        -----
+        This method is identical to `refresh` and exists to provide contextual clarity. Both methods produce identical results. When working with applications, natural language prefers "restart this content" instead of "refresh this content" since applications require a system process. When writing software that operates on multiple types of content (e.g., applications, documents, scripts, etc.), you may use either 'restart' or 'refresh' to achieve the same result.
```
nealrichardson, line 204: 1. I don't know that "restart" and "refresh" should be identical. I might think you have `restart()`, which works on interactive content (and errors on static), and `render()`, which works on static (and errors on interactive), and then `refresh()` is the generic. Then you have

```python
def refresh(self, **kwargs):
    if self.is_interactive():
        self.restart()
    else:
        self.render(**kwargs) # for variants/params
```

That way, if you want to be strict about what you're intending (restart or render), you can.

2. If you want to proceed in the current direction, you don't need _re_whatever(), just define one and have the other call it, you don't need them both to call some third function.

src/posit/connect/content.py
nealrichardson: We'll add a v1 version, but even so, you'll still have to fall back to the old APIs when you're on old (well, contemporary) Connect. This is fine for this PR, since I'm guessing you're not implementing all of the Variants features (I haven't read that far yet), but just wanted to put that out there for the future. It's been in production for years, people use it, so our notion of internal vs. public doesn't really matter.

toph-allen (approved):
Looks great! ✨

I don't have any specific feedback that should prevent merging — just running thoughts on stuff we've been discussing.

integration/tests/posit/connect/test_env.py
nealrichardson: Ah nm I see you have these too, just not in the integration tests 👍

src/posit/connect/env.py
nealrichardson: Ah cool so you do have these!

I'd add `keys()` as an alias to `find()`, as that seems like the more natural way to list the names of the env vars as dict.

`__getitem__` should probably exist but error with a message explaining that the values are encrypted so we can't show them. Likewise `items()` and similar.

nealrichardson (commented):
Cool! A few notes.

src/posit/connect/content.py
jonkeane: > The big caveat to this approach is upgrading the app_mode checks over time to accommodate new application modes.

We ran into [the same thing](https://github.com/rstudio/connectapi/pull/283#discussion_r1687197924) in connect_api. I wonder if we ought to have a more central place to retrieve this (either on the server, that enumerates what it knows about), or somewhere common in our code we could depend on. Of course, we don't need to / shouldn't solve that here, but it's not unique to this SDK!

src/posit/connect/content.py
```diff
@@ -137,12 +140,231 @@ class ContentItem(Resource):
...
+        url = urls.append(self.config.url, path)
+        response = self.session.post(url, json={"bundle_id": None})
+        result = response.json()
+        ts = tasks.Tasks(self.config, self.session)
+        return ts.get(result["task_id"])
+
+    def refresh(self) -> Task | None:
+        """Trigger a content refresh.
+
+        Submit a refresh request to the server for the content. After submission, the server executes an asynchronous process to refresh the content. This is useful when content is dependent on external information, such as a dataset.
```
jonkeane, line 177:
```suggestion
        Submit a refresh request to the server for the content. After submission, the server executes an asynchronous process to refresh the content. This is useful when content is dependent on external information, such as a dataset which has been updated.
```

src/posit/connect/content.py
tdstein: After discussing, we decided to implement `restart` and `render' and provide a guide error message if the content type does not support the invoked method.

PR #238, opened 2024-07-23
refactor: remove refresh and replace refresh with render

tdstein:
r:rrarrwr...

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -170,10 +170,10 @@ class ContentItem(Resource):
         ts = tasks.Tasks(self.config, self.session)
         return ts.get(result["task_id"])
 
-    def rerender(self) -> Task:
-        """Rerender the content.
+    def render(self) -> Task:
+        """Render the content.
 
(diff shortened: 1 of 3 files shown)


tdstein:
Here are the changes we discussed a little bit ago.

src/posit/connect/content.py
```diff
@@ -171,43 +170,80 @@ def deploy(self) -> tasks.Task:
         ts = tasks.Tasks(self.config, self.session)
         return ts.get(result["task_id"])
 
-    def refresh(self) -> Task | None:
-        """Trigger a content refresh.
+    def rerender(self) -> Task:
```
nealrichardson, line 173: Can this just be `render()`?

toph-allen (approved):
This looks good to me! Did you decide not to implement different subclasses for now? (I'm fine with that — this looks similar to how the R code currently looks — just curious!)

tdstein:
> This looks good to me! Did you decide not to implement different subclasses for now? (I'm fine with that. This looks similar to how the R code currently looks; I'm just curious!)

Yes, I decided not to for now. I implemented it and realized it wasn't worth the additional maintenance complexity. For what it's worth, I'm comfortable making a breaking change here in the future if we decide to do the abstraction.  At least until we get to a v1.0.0 release.

PR #239, opened 2024-07-24
refactor: context pattern

tdstein:
Replace standard dependency injection with the context pattern.

This has been on my mind for a while, and I needed to scratch my creativity itch. This simplifies the complexity of obtaining additional information throughout the resource stack. 

The existing implementation requires a large change every time we add a variable to the resource stack, such as passing the Connect version from the `Client` to the `Bundle`. Now that there is a standard code flow throughout the abstraction, migrating to a Context pattern makes sense. 

After this, we can add the Connect version to the `Context`, and then it will be available on the resource stack.

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -15,6 +15,7 @@ from .metrics import Metrics
 from .tasks import Tasks
 from .users import User, Users
 from .groups import Groups
+from .context import Context
 
 
 class Client:
@@ -155,12 +156,12 @@ class Client:
             if "url" in kwargs and isinstance(kwargs["url"], str):
                 url = kwargs["url"]
 
-        self.config = Config(api_key=api_key, url=url)
+        cfg = Config(api_key=api_key, url=url)
         session = Session()
-        session.auth = Auth(config=self.config)
+        session.auth = Auth(config=cfg)
         session.hooks["response"].append(hooks.check_for_deprecation_header)
         session.hooks["response"].append(hooks.handle_errors)
-        self.session = session
+        self.ctx = Context(api_key=cfg.api_key, session=session, url=cfg.url)
 
     @property
     def version(self) -> str:
@@ -184,7 +185,7 @@ class Client:
         User
             The currently authenticated user.
         """
-        return me.get(self.config, self.session)
+        return me.get(self.ctx)
 
     @property
     def oauth(self) -> OAuthIntegration:
@@ -196,7 +197,7 @@ class Client:
         OAuthIntegration
             The OAuth integration instance.
         """
-        return OAuthIntegration(config=self.config, session=self.session)
+        return OAuthIntegration(self.ctx)
 
     @property
     def groups(self) -> Groups:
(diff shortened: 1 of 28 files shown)


src/posit/connect/client.py
```diff
@@ -155,12 +156,12 @@ def __init__(self, *args, **kwargs) -> None:
...
 
-        self.config = Config(api_key=api_key, url=url)
+        cfg = Config(api_key=api_key, url=url)
         session = Session()
-        session.auth = Auth(config=self.config)
+        session.auth = Auth(config=cfg)
         session.hooks["response"].append(hooks.check_for_deprecation_header)
         session.hooks["response"].append(hooks.handle_errors)
-        self.session = session
+        self.ctx = Context(api_key=cfg.api_key, session=session, url=cfg.url)
```
nealrichardson, line 164: Another way to achieve this would be to subclass `requests.Session` and add the Config object as an attribute there, like:

```python
class ConnectSession(requests.Session):
    def __init__(self, connect_config):
        super().__init__()
        self.connect_config = connect_config
```

That way, I can still do `self.session.post()` etc. if I need to construct my own requests (as we do in some recipes) but still only have the one object to pass around. I'm not sure we would want to tell people to do `client.ctx.session.get()` to do a request, that seems unnecessarily complicated.

tdstein: The HTTP methods are still proxied through `Client` via `client.request`, `client.get`, `client.post`, etc.

https://github.com/posit-dev/posit-sdk-py/blob/cbc81cdf4f263ba3ba9d9f9972bf2335e86a9ff2/src/posit/connect/client.py#L302-L428

Are you referring to users of the posit-sdk or others working on this source code?

nealrichardson: There are a few cases when you need to make requests directly of the session and not using the Client because that way prepends the API URL. But right, not a common use case.

Nevertheless, what do you think about subclassing?

tdstein: Extending the `requests.Session` class seems odd to me strictly due to SoC. For example, housing the Connect version at `session.version` instead of `ctx.version` is strange since the version doesn't belong to the session. Maybe a better example is where to place a `logger`: `session.logger` or `ctx.logger`. In my experience, `ctx.logger` is typical.

Subclassing would be appropriate if we needed to modify or extend the `request.Session` behavior. But there isn't a need for that now.

I did some reading to refresh my understanding of the context design pattern. The original design addressed complex middleware chaining in server implementations. What I want to achieve is much simpler. It looks like the correct concept is simply parameter encapsulation. The correct approach is to change `Context` to `ResourceParameters` and only use it as part of the `Resource.__init__` and `Resources.__init__` method signatures.

To address the `client.ctx.session.get()` situation, we can revert to keeping the session as an attribute of the `Client` (e.g., `client.session.get()`). 

https://www.dre.vanderbilt.edu/~schmidt/PDF/Context-Object-Pattern.pdf

tdstein:
https://github.com/posit-dev/posit-sdk-py/pull/240

PR #241, opened 2024-07-25
refactor: decompose content module into a package

tdstein:
Decompose the content module into a package. Also renames Content to ContentResources and ContentItem to Content.

Since *content* is an [uncountable noun](https://dictionary.cambridge.org/grammar/british-grammar/content-or-contents), we are stuck choosing an alternative name for the singular or plural form. I use the singular content more frequently in both writing and speaking situations. The original decision was a *pick one* scenario. History aside, I think Content/ContentResources is more straightforward and more intuitive regardless.


src/posit/connect/content/contents.py
```diff
@@ -0,0 +1,287 @@
...
+from .. import urls
+from ..config import Config
+from ..resources import Resources
+from .content import Content
+
+
+class ContentResources(Resources):
+    """Content resources.
+
+    A resource for many pieces of content.
```
toph-allen, line 16: It's confusing to me that this class name is a plural, since this is a singular resource, no? 🤔

toph-allen (approved):
Few thoughts:

- The plural "content resources" confuses me. Aren't resources singular when talking about a certain type of object — you have a content resource that is the source for multiple items of content?
- `connectapi` uses the name `Content` as the class, although there is the constructor function `content_item()` just to keep things confusing! :D

tdstein:
> Few thoughts:
> 
> * The plural "content resources" confuses me. Aren't resources singular when talking about a certain type of object — you have a content resource that is the source for multiple items of content?
> * `connectapi` uses the name `Content` as the class, although there is the constructor function `content_item()` just to keep things confusing! :D

Hmm, let me see if I can clear things up. There are two base classes that represent the persistence layer of the stack: `Resource` and `Resources`. The plural version, `Resources`, is a pure repository pattern (or data access object) implementation whose job is to handle operations against the plural REST API endpoints (.e.g., `/bundles`).  The singular version, `Resource`, is both a repository pattern implementation and a data object. It is designed to interact with a singular REST API endpoint (e.g., `/bundles/1`).

In most cases, we can use the singular and plural forms of the noun to name the concrete implementations (i.e., `Bundle` and `Bundles`). But, for these pesky uncountable nouns (content), we must choose an alternate name for either the singular or the plural. This change attempts to address that by using the name `ContentResources` in the plural form and `Content` in the singular.

An alternative approach could be to use a more verbose naming system and to separate the singular repository and data representations. Example...

```python
from typing import List


class Resource(dict):
    pass


class ResourceRepository:
    def delete(self):
        pass

    def update(self):
        pass


class ResourcesRepository:
    def create(self) -> Resource:
        pass

    def find(self) -> List[Resource]:
        pass

    def get(self, uid: str) -> Resource | None:
        pass


class ContentResource(Resource):
    pass


class ContentResourceRepository(ResourceRepository):
    pass


class ContentResourcesRepository(ResourcesRepository):
    pass
```

PR #242, opened 2024-07-25
feat: add payload to ClientError and structure message

tdstein:


src/posit/connect/errors.py
```diff
@@ -1,15 +1,21 @@
...
-            f"{error_message} (Error Code: {error_code}, HTTP Status: {http_status} {http_message})"
-        )
+        message = {
+            "error_code": error_code,
+            "error_message": error_message,
+            "http_status": http_status,
+            "http_message": http_message,
+            "payload": payload,
+        }
+        super().__init__(json.dumps(message))
```
tdstein, line 21: Since the payload is sometimes a JSON object, it seems appropriate to structure the entire message. My initial implementation appended the payload to the end of the existing string, but I thought that looked odd.

example, `error message (Error Code: 0, HTTP Status: 404 Foo Bar) { 'payload': 'object' }`

dbkegley (approved):

dbkegley:
Thanks for jumping on this!

PR #243, opened 2024-07-26
feat: add content type helpers

tdstein:


jonkeane:
A description would be nice describing why. Also: do we have an issue in the connect (server) repo where we can discuss the possibility of having the server reply back with this split? We implement it here and in `connectapi`, and that's ok, but it does seem like something that the server should also be able to (self)-describe

tdstein:
> A description would be nice describing why. Also: do we have an issue in the connect (server) repo where we can discuss the possibility of having the server reply back with this split? We implement it here and in `connectapi`, and that's ok, but it does seem like something that the server should also be able to (self)-describe

Not that I could find. I created https://github.com/rstudio/connect/issues/27791

PR #244, opened 2024-07-26
refactor: add url class in place of urls module functions

tdstein:
Adds a new class, `urls.Url`, which wraps the existing `urls.create` and `urls.append` methods. This alleviates the need to `import  .urls` in most situations (see https://github.com/posit-dev/posit-sdk-py/pull/246).

The new class `urls.Url` supports Python append via the special method `__add__`. This is convenient for appending paths to the base endpoint. (e.g., `new_url = base_url + '/path'`).


jonkeane:
The changes look fine to me and slightly less boiler plate which is nice. Would you mind adding a description to the PR with pros / cons for the historical record?

toph-allen (approved):
😍

PR #245, opened 2024-07-26
build: reduce integration test suite startup time

tdstein:


Dockerfile
```diff
@@ -2,16 +2,13 @@ FROM python:3
 
 RUN apt-get update && apt-get install -y make
 
-WORKDIR /sdk
-
-COPY requirements.txt requirements-dev.txt vars.mk Makefile ./
+RUN --mount=type=cache,mode=0755,target=/root/.cache/pip pip install rsconnect
```
jonkeane, line 5: It feels a little strange to special case this one package. I presume it is the longest running install of all the other dependencies? Or?

tdstein: Yes, it has a lot of dependencies. But, I'm now realizing it should be `rsconnect-python`. `rsconnect` is the meta package for python, jupyter and pins. That seems a bit smaller, so I think it is safe to move back to requirements-dev.txt.

PR #247, opened 2024-07-26
build: add import formatting with autoflake and isort

tdstein:


Makefile
```diff
@@ -43,6 +43,8 @@ fix:
 
 fmt:
 	$(PYTHON) -m ruff format .
+	$(PYTHON) -m autoflake --remove-all-unused-imports --in-place --recursive .
+	$(PYTHON) -m isort .
```
mmarchetti, line 47: If you're using `ruff`, you can sort imports: https://docs.astral.sh/ruff/formatter/#sorting-imports

Makefile
```diff
@@ -38,10 +38,9 @@ dev:
 docs:
 	$(MAKE) -C ./docs
 
-fix:
-	$(PYTHON) -m ruff check --fix
-
 fmt:
+	$(PYTHON) -m autoflake --remove-all-unused-imports --in-place --recursive .
```
mmarchetti: `ruff check --fix` will also remove unused imports; if you only want to sort imports and remove unused, you can run `ruff check --select I,F401 --fix`. (`F401` is the code for unused imports, and `I` is the code group for isort-related issues).

tdstein: haha, I also just realized that. I had forgotten Ruff is a drop-in replacement for all of these tools.

toph-allen (approved):

CONTRIBUTING.md
```diff
@@ -22,7 +22,7 @@ Before contributing to the `posit-sdk`, ensure that the following prerequisites
 1. Create a new branch for your feature or bug fix.
 1. Run `make` to run the default development workflow.
 1. Make your changes and test them thoroughly using `make test`
-1. Run `make fmt`, `make lint`, and `make fix` to verify adherence to the project style guide.
+1. Run `make fmt` and `make lint` to verify adherence to the project style guide.
 1. Commit your changes and push them to your forked repository.
 1. Submit a pull request to the main repository.
```
jonkeane, line 28: I can't comment lower where it's talked about, but do we give folks steps for setting up ruff? IMO having a "run these lines of code and it will make your dev environment work" really helpful. 

Alternatively / in addition: this is something we could trigger a bot on a comment and then CI deals with having an env setup. I've found this pattern to be suuuuper helpful for drop-by contributors

tdstein: > IME, this works sometimes, but not always.

Sure. We can add more context if needed in the future. I believe there are GitHub Actions that will annotate the source code with lining errors.

PR #248, opened 2024-07-26
chore: apply flake8 builtin rules

tdstein:
Adds formatting rule and fixes variable names that shadow Python builtins.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ docstring-code-format = true
 docstring-code-line-length = "dynamic"
 
 [tool.ruff.lint]
-select = ["D", "F401", "I"]
+select = ["A", "D", "F401", "I"]
 ignore = [
     # NumPy style docstring convention with noted exceptions.
     # https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,32 @@ docstring-code-format = true
 docstring-code-line-length = "dynamic"
 
 [tool.ruff.lint]
-select = ["A", "D", "F401", "I"]
+select = [
+    # flake8-builtins
+    # https://docs.astral.sh/ruff/rules/#flake8-builtins-a
+    #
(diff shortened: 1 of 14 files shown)


toph-allen (approved):

pyproject.toml
```diff
@@ -44,7 +44,7 @@ docstring-code-format = true
 docstring-code-line-length = "dynamic"
 
 [tool.ruff.lint]
-select = ["D", "F401", "I"]
+select = ["A", "D", "F401", "I"]
```
jonkeane, line 47: I should probably RTFM, but what's "A" here? I know the other part of this list are already here, but would it make sense to add a comment about what these are?

src/posit/connect/bundles.py
```diff
@@ -235,16 +235,16 @@ def __init__(
         super().__init__(params)
         self.content_guid = content_guid
 
-    def create(self, input: io.BufferedReader | bytes | str) -> Bundle:
+    def create(self, archive: io.BufferedReader | bytes | str) -> Bundle:
```
jonkeane, line 238: Is this to avoid using `input` as an argument?

pyproject.toml
tdstein: https://docs.astral.sh/ruff/rules/#flake8-builtins-a

I'll add some comments to help out the next person.

src/posit/connect/bundles.py
tdstein: Correct. Since `input` is a builtin function.

jonkeane: We might not need to worry cause this is internal enough or no one is yet using it, but does this change have user-facing implications?

tdstein: Yes, it does if anyone is using kwargs when calling this method. Since it's early and we are pre-1.0.0, I am ok releasing this under 0.4.0 along with documentation of the breaking changes.

tdstein: Yes, I think release notes are the correct place to write them.

tdstein: I started a migration guide in the changelog: https://github.com/posit-dev/posit-sdk-py/releases

PR #251, opened 2024-07-31
feat: replace restart environment variable hash with unix epoch

tdstein:
The hash was introduced to reduce the chance of collisions with existing environment variables introduced by the user or another restart invocation. A UNIX epoch provides similar collision protection while providing additional metadata for debugging if the unset routine fails.


toph-allen (approved):
Looks good!

PR #252, opened 2024-08-06
feat!: improve compatibility with databricks sql client

dbkegley:
- Improves compatibility/usability between the Databricks python SDK and the Databricks SQL client credentials_provider interfaces: https://github.com/databricks/databricks-sql-python/issues/148#issuecomment-2271561365
- Updates all databricks viewer-auth examples to refer to the next posit-sdk release (0.4.0) in their requirements.txt

For reviewers: The important changes relevant for this PR's functionality are in `src/posit/connect/external/databricks.py`

The new credentials_strategy takes a fallback option that will be used automatically during local development, this way content can run locally and on Connect w/o requiring any code changes.

```python
import streamlit as st
from databricks import sql
from posit.connect.external.databricks import PositCredentialsStrategy

session_token = st.context.headers.get("Posit-Connect-User-Session-Token")
posit_strategy = PositCredentialsStrategy(
    local_strategy=databricks_cli,
    user_session_token=session_token)
cfg = Config(
    host=DATABRICKS_HOST_URL,
    # uses Posit's custom credential_strategy if running on Connect,
    # otherwise falls back to the strategy defined by local_strategy
    credentials_strategy=posit_strategy)

databricks_user = CurrentUserAPI(ApiClient(cfg)).me()
st.write(f"Hello, {databricks_user.display_name}!")

with sql.connect(
    server_hostname=DATABRICKS_HOST,
    http_path=SQL_HTTP_PATH,
    # https://github.com/databricks/databricks-sql-python/issues/148#issuecomment-2271561365
    credentials_provider=posit_strategy.sql_credentials_provider(cfg)
) as connection:
    with connection.cursor() as cursor:
        cursor.execute("SELECT * FROM samples.nyctaxi.trips LIMIT 10;")
        result = cursor.fetchall()
```

### DIFF
--- a/src/posit/connect/external/databricks.py
+++ b/src/posit/connect/external/databricks.py
@@ -5,73 +5,108 @@ from typing import Callable, Dict, Optional
 from ..client import Client
 from ..oauth import OAuthIntegration
 
-HeaderFactory = Callable[[], Dict[str, str]]
+"""
+NOTE: These APIs are provided as a convenience and are subject to breaking changes:
+https://github.com/databricks/databricks-sdk-py#interface-stability
+"""
 
+# The Databricks SDK CredentialsProvider == Databricks SQL HeaderFactory
+CredentialsProvider = Callable[[], Dict[str, str]]
 
-# https://github.com/databricks/databricks-sdk-py/blob/v0.20.0/databricks/sdk/credentials_provider.py
-# https://github.com/databricks/databricks-sql-python/blob/v3.1.0/src/databricks/sql/auth/authenticators.py
-# In order to keep compatibility with the Databricks SDK
-class CredentialsProvider(abc.ABC):
-    """Protocol Databricks authentication.
+class CredentialsStrategy(abc.ABC):
+    """Maintain compatibility with the Databricks SQL/SDK client libraries.
 
-    A call-side interface for the Databricks Rest API.
+    https://github.com/databricks/databricks-sql-python/blob/v3.3.0/src/databricks/sql/auth/authenticators.py#L19-L33
+    https://github.com/databricks/databricks-sdk-py/blob/v0.29.0/databricks/sdk/credentials_provider.py#L44-L54
     """
-
     @abc.abstractmethod
     def auth_type(self) -> str:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def __call__(self, *args, **kwargs) -> HeaderFactory:
+    def __call__(self, *args, **kwargs) -> CredentialsProvider:
         raise NotImplementedError
 
 
-class PositOAuthIntegrationCredentialsProvider(CredentialsProvider):
+def _is_local() -> bool:
+    """Returns true if called from a piece of content running on a Connect server.
+
+    The connect server will always set the environment variable `RSTUDIO_PRODUCT=CONNECT`.
+    We can use this environment variable to determine if the content is running locally
+    or on a Connect server.
+    """
+    return not os.getenv("RSTUDIO_PRODUCT") == "CONNECT"
+
+
+class PositCredentialsProvider:
     def __init__(self, posit_oauth: OAuthIntegration, user_session_token: str):
         self.posit_oauth = posit_oauth
         self.user_session_token = user_session_token
 
+    def __call__(self) -> Dict[str, str]:
+        access_token = self.posit_oauth.get_credentials(
+            self.user_session_token
+        )["access_token"]
+        return {"Authorization": f"Bearer {access_token}"}
+
+
+class PositCredentialsStrategy(CredentialsStrategy):
+
+    def __init__(self,
+        local_strategy: CredentialsStrategy,
+        user_session_token: Optional[str] = None,
+        client: Optional[Client] = None
+    ):
+        self.user_session_token = user_session_token
+        self.local_strategy = local_strategy
+        self.client = client
+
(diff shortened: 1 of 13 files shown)


src/posit/connect/external/databricks.py
```diff
@@ -5,73 +5,89 @@
...
 
     @abc.abstractmethod
-    def __call__(self, *args, **kwargs) -> HeaderFactory:
+    def __call__(self, *args, **kwargs) -> CredentialsProvider:
         raise NotImplementedError
 
 
-class PositOAuthIntegrationCredentialsProvider(CredentialsProvider):
+# Use this environment variable to determine if the
+# client SDK was initialized from a piece of content running on a Connect server.
```
tdstein, line 30: Could you move this to a docstring?

src/posit/connect/external/databricks.py
```diff
@@ -5,73 +5,89 @@
...
-    # If the user-session-token wasn't provided and we're running on Connect then we raise an exception.
-    # user_session_token is required to impersonate the viewer.
-    if user_session_token is None:
-        raise ValueError(
-            "The user-session-token is required for viewer authentication."
+    def auth_type(self) -> str:
+        if _is_local():
+            return self.local_strategy.auth_type()
+        else:
+            return "posit-oauth-integration"
```
tdstein, line 74: Could you add a docstring explaining this and why it works?

tdstein:
You can run `make lint` locally to debug the linting issues.

tdstein (commented):
Changes look reasonable. Is it worth adding any unit tests along with this change?

src/posit/connect/external/databricks.py
```diff
@@ -5,73 +5,89 @@
...
+            self.user_session_token
+        )["access_token"]
+        return {"Authorization": f"Bearer {access_token}"}
 
 
-# Use this environment variable to determine if the
-# client SDK was initialized from a piece of content running on a Connect server.
-def is_local() -> bool:
-    return not os.getenv("RSTUDIO_PRODUCT") == "CONNECT"
+class PositOAuthIntegrationCredentialsStrategy(CredentialsStrategy):
```
nealrichardson, line 47: I know this is superficial, but do we have options for making this class name shorter? As a user, giant names like this scare me and make me assume that the code is super complex. 

Would `PositCredentialsStrategy` work?

tdstein: Same here! That or `ConnectCredentialsStrategy`.

zackverham: Just another option in the merry-go-round of word permutations - because this is already under the `posit/connect` directory, would it be more informative to call it an `OAuthCredentialsStrategy`? Then we could have other credentials strategies that also work for posit products / connect but we don't run into naming collisions.

dbkegley: I opted to go with `PositCredentialsStrategy` to avoid a naming conflict with the databricks sdk's `OAuthCredentialsStrategy`.  I think this is a good balance of brevity and descriptive-ness since we are still under the `posit/connect` namespace 

https://github.com/databricks/databricks-sdk-py/blob/7d22b4d3727478f0f5dbeb34b7f6fc17a03e31b7/databricks/sdk/credentials_provider.py#L57

tests/posit/connect/external/test_databricks.py
```diff
@@ -0,0 +1,71 @@
...
+        client = Client(api_key="12345", url="https://connect.example/")
+        cs = PositCredentialsStrategy(local_strategy=local_strategy(),
+                                      user_session_token="cit",
+                                      client=client)
+        cp = cs()
+        assert cs.auth_type() == "local"
+        assert cp() == {"Authorization": "Bearer static-pat-token"}
+
+        # posit_strategy is used when the content is running on Connect
+        os.environ["RSTUDIO_PRODUCT"] = "CONNECT"
```
tdstein, line 63: It's possible to mock the environment variables using `unittest.mock.patch`: https://github.com/posit-dev/posit-sdk-py/blob/8648f8038f717514652796634c3804a024642ce2/tests/posit/connect/test_config.py#L7

This avoids modifying the developer's environment entirely.

src/posit/connect/external/databricks.py
```diff
@@ -5,73 +5,109 @@
...
+
+        # If the user-session-token wasn't provided and we're running on Connect then we raise an exception.
+        # user_session_token is required to impersonate the viewer.
+        if self.user_session_token is None:
+            raise ValueError(
+                "The user-session-token is required for viewer authentication."
+            )
+
+        if self.client is None:
+            self.client = Client()
```
tdstein, line 109: Could you help me understand the situation in which self.client is None?  My concern here if someone needs to set the endpoint or api_key directory instead of using the environment variables.

dbkegley: `self.client` might be None when the content is running locally in which case the user can't use the `PositCredentialsStrategy`.  We would fallback to the `local_strategy` instead. When the content is running on Connect then we know that Connect will set `CONNECT_API_KEY` and `CONNECT_SERVER_URL` for the content so the `Client()` would be allowed to initialize

tdstein: Got it. Thank you!

tdstein (approved):

tdstein (approved):
Everything looks good to me. Thanks for the additional documentation, tests, and explanations!

PR #256, opened 2024-08-08
fix: use correct args and kwargs definitions

tdstein:
Various methods are defined with improper variadic positional and keyword arguments. The original intention for `@overload` signatures was to provide type support for pass-throughs to API query parameters or JSON arguments. Along the way, *args was introduced to these methods, assuming we wanted to match the dict function signature. This isn't very clear in practice since passing SupportsKeysAndGetItem typed variables is rare. Therefore, the *args part of the function signature is removed wherever possible. In the update(self, *args, **kwargs), we cannot remove *args since we inherit from dict, and dict enforces this method signature.

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -229,6 +229,7 @@ class ContentItem(Resource):
     @overload
     def update(
         self,
+        *args,
         name: str = ...,
         title: Optional[str] = ...,
         description: str = ...,
@@ -254,6 +255,7 @@ class ContentItem(Resource):
         default_r_environment_management: Optional[bool] = ...,
         default_py_environment_management: Optional[bool] = ...,
         service_account_name: Optional[str] = ...,
+        **kwargs,
     ) -> None:
         """Update the content item.
 
@@ -578,6 +580,7 @@ class Content(Resources):
     @overload
     def create(
         self,
+        *,
         name: str = ...,
         title: Optional[str] = ...,
         description: str = ...,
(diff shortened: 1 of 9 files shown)


src/posit/connect/content.py
```diff
@@ -229,6 +229,7 @@ def restart(self) -> None:
     @overload
     def update(
         self,
+        *args,
```
zackverham, line 232: are these *args additions just for forwards / backwards compatibility?

src/posit/connect/metrics/visits.py
```diff
@@ -153,6 +153,7 @@ def find(self, *args, **kwargs) -> List[VisitEvent]:
     @overload
     def find_one(
         self,
+        *,
```
zackverham, line 156: same question about the places we're adding `*`

zackverham (approved):

src/posit/connect/content.py
tdstein: These are required to match the dict.update signature. Since Resource inherits from dict, we have to adhere to it. You can see this by removing *args and running `make lint`.

tdstein: See below first. This is the equivalent to *, except in this case we have to include *args to match the update signature.

src/posit/connect/metrics/visits.py
tdstein: This is to say, all arguments after `*` must be provided as keyword arguments. So, the user must call `find_one(content_guid="asdf")`. `find_one("asdf")` will throw an error.

tdstein: This is what we want since positional argument can change over time.

PR #259, opened 2024-08-27
fix: correct type signatures according to pyright

tdstein:
This change fixes errors and warnings issued by Pyright. 

The update method is added to the base Resource class, which fixes type-checking issues against MutableMapping.

### DIFF
--- a/src/posit/connect/resources.py
+++ b/src/posit/connect/resources.py
@@ -35,6 +35,9 @@ class Resource(ABC, dict):
     def __setattr__(self, name: str, value: Any) -> None:
         raise AttributeError("cannot set attributes: use update() instead")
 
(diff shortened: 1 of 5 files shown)


src/posit/connect/resources.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, params: ResourceParameters, **kwargs):
     def __setattr__(self, name: str, value: Any) -> None:
         raise AttributeError("cannot set attributes: use update() instead")
 
+    def update(self, *args, **kwargs):
+        super().update(*args, **kwargs)
+
```
tdstein, line 40: @zackverham - this resolves the overload signature issue you encountered.

zackverham (approved):

PR #261, opened 2024-09-03
chore: maintenance tasks

tdstein:
Accomplishes a few maintenance tasks:
- Fix `python` binary configuration when using `uv` in vars.mk.
- Change `make lint` configuration to target Python files only and remove unnecessary exclude configuration.
- Fix `__init__.py` export to use `as` naming syntax (`as connect`).
- Set classifier 'Development Status :: 4 - Beta'.

### DIFF
--- a/src/posit/__init__.py
+++ b/src/posit/__init__.py
@@ -1,7 +1,3 @@
 """The Posit SDK."""
 
-from . import connect  # noqa
-
-__all__ = [
-    "connect"
-]
+from . import connect as connect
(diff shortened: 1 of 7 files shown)


src/posit/__init__.py
```diff
@@ -1,7 +1,5 @@
 """The Posit SDK."""
 
-from . import connect  # noqa
+from . import connect as connect
```
toph-allen, line 3: What does `import connect as connect` do?

toph-allen (approved):

src/posit/__init__.py
tdstein: https://docs.astral.sh/ruff/rules/unused-import/

But, I now see I can also remove the __all__ statement. Going to update that now.

PR #265, opened 2024-09-05
build: always use uv for dependency management

tdstein:
Modifies the build tooling to always use `uv pip` instead of `pip`.  

Additional checks are added to ensure that `uv` exists on the system. If `uv` does not exists, then it is installed via `pip`. If `pip` is not installed, it is installed via `python -m ensurepip`.  The `pip` check is required for a virtual environment created by `uv venv` since by default, `uv venv` does not install `pip`.

This reduces the execution time of `make` from 22s to 5s on my machine without a Pip cache. With a hydrated PIp cache, the execution time is reduced from 8s to 4s.


zackverham (approved):
cool - makes sense to me! 

I'm not super familiar with `uv` but it looks to be pretty easy to drop-in as a replacement.

PR #266, opened 2024-09-05
feat: deprecates field access by attribute.

tdstein:
Remove all `@property` annotated methods used for direct field access on concrete resources. To maintain backward compatibility, a new `__getattr__` implementation is added to Resource. This implementation invokes `__getitem__` when the key exists in self. A noisy deprecation warning alerts users that this functionality will be removed in v1.0.0.

### DIFF
--- a/src/posit/connect/bundles.py
+++ b/src/posit/connect/bundles.py
@@ -9,130 +9,10 @@ from . import resources, tasks
 
 
 class BundleMetadata(resources.Resource):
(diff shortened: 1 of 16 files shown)


toph-allen (approved):
How does this relate to #271 in terms of scheduling / release?

tdstein:
@zackverham, this includes changes to the OAuth resources you added yesterday.

zackverham (approved):

PR #268, opened 2024-09-05
feat!: Implement snowflake auth helpers

### LINKED ISSUE #269: Implement snowflake helpers

dbkegley:
- Reorganize examples into `databricks` and `snowflake` sub-directories
- Add `external.snowflake` helpers
- Add snowflake streamlit example
- Move `_is_local` to `__init__.py`

Breaking changes:
- `posit.connect.external.PositCredentialsProvider.__init__()` now accepts a `Client` instead of an `OAuthIntegration` resource.
- All fields of `posit.connect.external.PositCredentialsProvider` and `posit.connect.external.PositCredentialsStrategy` are now `_internal`
- Change order of named kwargs to `posit.connect.external.PositCredentialsStrategy.__init__()`. Only breaking for callers who use `*args` instead of `**kwargs`

closes #269


zackverham:
not sure if this has come up anywhere else - but currently we don't indicate w/ the directory structure that our examples are specifically for the oauth integration pieces of the SDK. As we're breaking out sub-directories for the type of integration (snowflake vs databricks) I wonder if we need to do more to flag exactly what these are examples of?

tdstein:
> not sure if this has come up anywhere else - but currently we don't indicate w/ the directory structure that our examples are specifically for the oauth integration pieces of the SDK. As we're breaking out sub-directories for the type of integration (snowflake vs databricks) I wonder if we need to do more to flag exactly what these are examples of? 

These are well suited for cookbook recipes or extensions as well.

dbkegley:
> These are well suited for cookbook recipes or extensions as well.

I'm happy to start moving these over to cookbooks before the Sept Connect release. I think we should have at least 1 release where they exist in both places and the README in the examples dir should start pointing folks toward the cookbooks ASAP. Next Connect release we can remove the examples from the SDK and just maintain the cookbooks going forward.

tdstein:
> > These are well suited for cookbook recipes or extensions as well.
> 
> I'm happy to start moving these over to cookbooks before the Sept Connect release. I think we should have at least 1 release where they exist in both places and the README in the examples dir should start pointing folks toward the cookbooks ASAP. Next Connect release we can remove the examples from the SDK and just maintain the cookbooks going forward.

Awesome. Yes, I agree. It's best to hold off on deleting them until the recipes make it to docs.posit.co.

src/posit/connect/external/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+
+def _is_local() -> bool:
+    """Returns true if called from a piece of content running on a Connect server.
+
+    The connect server will always set the environment variable `RSTUDIO_PRODUCT=CONNECT`.
+    We can use this environment variable to determine if the content is running locally
+    or on a Connect server.
+    """
+    return not os.getenv("RSTUDIO_PRODUCT") == "CONNECT"
```
tdstein, line 14: I haven't found a clear answer to best practices, but I've recently started reducing __init__.py to only include named exports. Then placing any shared module code in a file of the same name; `external/external.py` here.

This also eliminates the need to mark the method as private since it won't be declared in the __init__.py file.

src/posit/connect/external/snowflake.py
```diff
@@ -0,0 +1,44 @@
...
+        self,
+        local_authenticator: Optional[str] = None,
+        client: Optional[Client] = None,
+        user_session_token: Optional[str] = None,
+    ):
+        self._local_authenticator = local_authenticator
+        self._client = client
+        self._user_session_token = user_session_token
+
+    def authenticator(self) -> Optional[str]:
```
tdstein, line 21: Should this be a `@property`?

src/posit/connect/external/snowflake.py
```diff
@@ -0,0 +1,44 @@
+from typing import Optional
+
+from . import _is_local
+from ..client import Client
+
+"""
+NOTE: The APIs in this module are provided as a convenience and are subject to breaking changes.
+"""
+
+class PositAuthenticator:
```
tdstein, line 10: Is this intended to work across Posit products? Would PositConnectAuthentication be more apt? Or even `Authenticator`?


Also, should we consider changing external to integrations? This seems fairly self-documenting...

```py
from posit.connect.integrations.snowflake import Authenticator

auth = Authenticator()
```

src/posit/connect/external/databricks.py
```diff
@@ -41,13 +40,13 @@ def _is_local() -> bool:
 
 
 class PositCredentialsProvider:
-    def __init__(self, posit_oauth: OAuthIntegration, user_session_token: str):
-        self.posit_oauth = posit_oauth
-        self.user_session_token = user_session_token
+    def __init__(self, client: Client, user_session_token: str):
+        self._client = client
+        self._user_session_token = user_session_token
```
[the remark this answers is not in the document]
tdstein: Thanks! If you can capture the breaking changes in the PR description, it will be easy to pull into the release notes.

tdstein (commented):
Looks great overall! I have a few minor questions and comments.

src/posit/connect/external/snowflake.py
dbkegley: I was following the naming style that we landed on here: https://github.com/posit-dev/posit-sdk-py/pull/252#discussion_r1705997523

> Also, should we consider changing external to integrations? This seems fairly self-documenting...

Since these APIs are still evolving I like having them under the `external` namespace for now. Maybe we consider re-naming and moving under the `integrations` namespace once we're more confident there wont be further breaking changes?

tdstein: Thanks! At least I'm consistent... 😛

tdstein (approved):

PR #271, opened 2024-09-06
refactor!: drop support field access by attribute. 

### LINKED ISSUE #262: Remove `@property` implementations for field access
**Note: this change requires updating the cookbook documentation. issue forthcoming...**

The initial intention of using the `@property` annotation for field access was to provide compatibility between Connect versions without having to modify code using posit-sdk. Since then, I've yet to encounter a field whose name has changed between Connect versions. Therefore implementing property methods has introduced a lot of cruft in the codebase. Additionally, we often have to mark getters with `# type: ignore` to ensure compatibility, which is counterintuitive.

An alternative solution to compatibility between Connect versions is to override the __getitem__ and __setitem__ methods as shown below. This is a simple approach and much more concise than our current implementation.

```py
import warnings


class CompatibleDict(dict):
    deprecated_keys = {"old": "new"}

    def __getitem__(self, key):
        # Check if the key is deprecated and use the new key if necessary
        if key in self.deprecated_keys:
            new_key = self.deprecated_keys[key]
            print(f"Warning: '{key}' is deprecated, use '{new_key}' instead.")
            key = new_key
        return super().__getitem__(key)

    def __setitem__(self, key, value):
        if key in self.deprecated_keys:
            new_key = self.deprecated_keys[key]
            warnings.warn(f"Warning: '{key}' is deprecated, use '{new_key}' instead.")
            key = new_key
        super().__setitem__(key, value)
```

Since this is breaking change, I propose a two-phased approach..

*Phase 1* (next minor release e.g., v.0.5.0)

Remove all property methods for field access and implement __getattr__  and __setattr__ on `Resource` as follows. This implementation should also include a deprecation warning.

```py
    def __getattr__(self, name):
        # Check if the name exists as a key in the dictionary
        if name in self:
            return self[name]
        # If the name does not exist as a key, raise the AttributeError
        raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")

    # Optional: __setattr__ to support setting attributes
    def __setattr__(self, name, value):
        self[name] = value
```

*Phase 2* (next major release e.g., v1.0.0)

Remove the __getattr__ and __setattr__ implementations defined above.

tdstein:
BREAKING CHANGE: removes support for accessing resource fields using
`__getattr__` (e.g., `user.username`). Use `__getitem__` instead (e.g.,
`user["username"]`.

depends on #265 and #266
closes #262

### DIFF
--- a/integration/tests/posit/connect/test_content.py
+++ b/integration/tests/posit/connect/test_content.py
@@ -23,7 +23,7 @@ class TestContent:
         assert self.client.content.count() == 1
 
     def test_get(self):
-        assert self.client.content.get(self.content.guid) == self.content
+        assert self.client.content.get(self.content["guid"]) == self.content
 
     def test_find(self):
         assert self.client.content.find()
@@ -34,12 +34,12 @@ class TestContent:
     def test_content_item_owner(self):
         item = self.client.content.find_one(include=None)
         owner = item.owner
-        assert owner.guid == self.client.me.guid
+        assert owner["guid"] == self.client.me["guid"]
 
     def test_content_item_owner_from_include(self):
         item = self.client.content.find_one(include="owner")
         owner = item.owner
-        assert owner.guid == self.client.me.guid
+        assert owner["guid"] == self.client.me["guid"]
 
     @pytest.mark.skipif(
         CONNECT_VERSION <= version.parse("2024.04.1"),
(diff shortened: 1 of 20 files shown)


toph-allen (approved):
That's a lot of code deleted! Definitely a worthwhile trade-off. :)

PR #286, opened 2024-09-11
feat: Implements OAuth association API resource

zackverham:
Adds support for requesting and updating oauth associations through the Connect API.

### DIFF
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -7,6 +7,8 @@ import time
 from posixpath import dirname
 from typing import Any, List, Optional, overload
 
+from posit.connect.oauth.associations import ContentItemAssociations  
+
 from . import tasks
 from .bundles import Bundles
 from .env import EnvVars
@@ -26,6 +28,13 @@ class ContentItem(Resource):
         if key == "owner" and isinstance(v, dict):
             return ContentItemOwner(params=self.params, **v)
         return v
+
+    @property
+    def oauth_associations(self) -> ContentItemAssociations:
+        if self.guid is None:
+            raise ValueError("ContentItem must have a guid to have associations")
+        return ContentItemAssociations(self.params, content_guid=self.guid)
+
 
     def delete(self) -> None:
         """Delete the content item."""
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -7,7 +7,7 @@ import time
 from posixpath import dirname
 from typing import Any, List, Optional, overload
 
-from posit.connect.oauth.associations import ContentItemAssociations  
+from posit.connect.oauth.associations import ContentItemAssociations
 
 from . import tasks
 from .bundles import Bundles
@@ -16,6 +16,18 @@ from .permissions import Permissions
 from .resources import Resource, ResourceParameters, Resources
 from .tasks import Task
 from .variants import Variants
+
+
+class ContentItemOAuth(Resource):
+    def __init__(self, params: ResourceParameters, content_guid: str) -> None:
+        super().__init__(params)
+        self.content_guid = content_guid
+
+    @property
+    def associations(self) -> ContentItemAssociations:
+        return ContentItemAssociations(
+            self.params, content_guid=self.content_guid
+        )
 
 
 class ContentItemOwner(Resource):
@@ -30,11 +42,10 @@ class ContentItem(Resource):
         return v
 
     @property
-    def oauth_associations(self) -> ContentItemAssociations:
-        if self.guid is None:
-            raise ValueError("ContentItem must have a guid to have associations")
-        return ContentItemAssociations(self.params, content_guid=self.guid)
-
+    def oauth(self) -> ContentItemOAuth:
+        if "guid" not in self:
+            raise ValueError("ContentItemOAuth requires content guid")
+        return ContentItemOAuth(self.params, content_guid=self["guid"])
 
     def delete(self) -> None:
         """Delete the content item."""
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -43,8 +43,6 @@ class ContentItem(Resource):
 
     @property
     def oauth(self) -> ContentItemOAuth:
-        if "guid" not in self:
-            raise ValueError("ContentItemOAuth requires content guid")
         return ContentItemOAuth(self.params, content_guid=self["guid"])
 
     def delete(self) -> None:
--- a/src/posit/connect/oauth/associations.py
+++ b/src/posit/connect/oauth/associations.py
@@ -1,0 +1,77 @@
+"""OAuth association resources."""
+
+from typing import List 
+
+from ..resources import Resource, ResourceParameters, Resources
+
+
+class Association(Resource):
+    pass
+
+
+class IntegrationAssociations(Resources):
+    """IntegrationAssociations resource."""
+
+    def __init__(self, params: ResourceParameters, integration_guid: str) -> None:
+        super().__init__(params)
+        self.integration_guid = integration_guid 
+
+    def find(self) -> List[Association]:
+        """Find OAuth associations.
+
+        Returns
+        -------
+        List[Association]
+        """
+        path = f"v1/oauth/integrations/{self.integration_guid}/associations"
+        url = self.params.url + path
+
+        response = self.params.session.get(url)
+        return [
+            Association(
+                self.params,
+                **result,
+            )
+            for result in response.json()
+        ]
+
+
+class ContentItemAssociations(Resources):
+    """ContentItemAssociations resource."""
+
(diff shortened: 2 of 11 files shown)


src/posit/connect/oauth/associations.py
```diff
@@ -0,0 +1,77 @@
...
+        return [
+            Association(
+                self.params,
+                **result,
+            )
+            for result in response.json()
+        ]
+
+
+    def update(self, integration_guid: None | str) -> None:
```
zackverham, line 66: @dbkegley @tdstein would love thoughts on the best way to represent this API interaction through the SDK. I don't think this is precisely what we want, but its a little ambiguous. 

The actual API endpoint accepts a list of json objects - this is designed with future flexibility in mind, but for now, the only valid inputs are: 

[{"oauth_integration_guid": <guid>}] to associate a content item with an integration
[] to dissociate the content item with all integrations

Given that currently you can only set one association per content item, I think it makes sense to support a function signature that accepts a single integration guid outside of a list. This breaks the semantics of PUTting an empty list to dissociate the content from integrations, though. 

I put `None` in as a placeholder for now to facilitate discussion, but I don't really like it all that much. I'm not sure what a better alternative would be.

src/posit/connect/content.py
```diff
@@ -27,6 +29,13 @@ def __getitem__(self, key: Any) -> Any:
             return ContentItemOwner(params=self.params, **v)
         return v
 
+    @property
+    def oauth_associations(self) -> ContentItemAssociations:
+        if self.guid is None:
+            raise ValueError("ContentItem must have a guid to have associations")
```
tdstein, line 35: You should do this check as part of `ContentItemAssociation.__init__`.

Also, `self["guid"]` instead of `self.guid`. See https://github.com/posit-dev/posit-sdk-py/pull/271

src/posit/connect/oauth/associations.py
tdstein: It feels like the correct expression is `content.associations.create(integration_guid)`, which should return a simple object that maintains the integration_guid and provides a `delete` method. The delete method would need to remove the association from the list.

```
class Association(Resource):
    def delete(self) -> None:
        # call `v1/content/{self.content_guid}/oauth/integrations/association` to remove the association.
```

fwiw, here is how I am thinking about the REST API implementation in my head, even though I know this doesn't represent the actual implementation.

Create an association
```
POST .../oauth/integrations/associations
{
    'integration_guid': 'whatever..."
}

Response:
{
    'guid': 'the-association-guid',
    'integration_guid': 'whatever...',
    'content_guid': 'the-content-guid'
}
```

Delete an association
```
DELETE .../the-content-guid/oauth/integrations/associations/the-association-guid
```

src/posit/connect/content.py
```diff
@@ -27,6 +29,13 @@ def __getitem__(self, key: Any) -> Any:
             return ContentItemOwner(params=self.params, **v)
         return v
 
+    @property
+    def oauth_associations(self) -> ContentItemAssociations:
```
tdstein, line 33: Maybe this should be expressed as `content.oauth.associations`?

src/posit/connect/oauth/associations.py
zackverham: in the current API implementation, you can't create / delete singular oauth associations - there aren't public CRUD operations that let you use an association in isolation. So I don't think your proposed pattern would be possible. 

You can only PUT a list of oauth integration guids to the`<content_guid>oauth/integrations/associations` endpoint - any PUTs completely wipe the current set of associations and replace it with a new set of associations.

src/posit/connect/content.py
zackverham: I like that alot better, yeah. Would the oauth property just be a passthrough class that we hang the associations resource off of?

src/posit/connect/content.py
zackverham: doing the check inside of the `__init__` does produce slightly rough behavior, i.e.

```

    @property
    def associations(self) -> IntegrationAssociations:
        return IntegrationAssociations(self.params, integration_guid=self["guid"])


```

if `self["guid"]` is unset, we will get a `KeyError` before we get into the constructor.

tdstein: Ah, I see. Disregard that.

src/posit/connect/oauth/associations.py
tdstein: I see. When support for multiple associations is added, will the PUT operation still be an overwrite? If so, how should multiple people coordinate the modification when they each want to add a single association to the list simultaneously?

zackverham: its a good question. I don't know if we have an answer to this - currently, the asnwer is "whoever PUTs last wins." 

Really it should only be publishers / admins who use this endpoint to interact with content that they own, so I wouldn't expect there to be many use cases where there's parallel requests coming from many clients. We really built the API with the expected client primarily being the dashboard.

zackverham:
# Test Notes

Here's a test script illustrating / manually validating the behavior of the associations resources: 

```
from posit.connect.client import Client
import json


c = Client("http://localhost:3939", "<redacted>")

integration_a_guid = "22644575-a27b-4118-ad06-e24459b05126"
integration_b_guid = "9dd7ef9d-ccd4-46f0-b76c-12b00f1ff594"
content_guid = "9dfa33ee-62ea-4a69-8b1a-51735c70dba9"

# get content item
content_item = c.content.get(content_guid)

# set initial state for testing
content_item.oauth.associations.update(integration_a_guid)

# associated with integration_a
association_list = content_item.oauth.associations.find()
print(json.dumps(association_list))

# Output:
# [{"app_guid": "9dfa33ee-62ea-4a69-8b1a-51735c70dba9", "oauth_integration_guid": "22644575-a27b-4118-ad06-e24459b05126", "oauth_integration_name": "keycloak", "oauth_integration_description": "", "oauth_integration_template": "custom", "created_time": "2024-09-12T19:05:11Z"}]

# clear associations
content_item.oauth.associations.delete()
association_list = content_item.oauth.associations.find()
print(json.dumps(association_list))

# Output: 
# []

# associated with integration_b
content_item.oauth.associations.update(integration_b_guid)
association_list = content_item.oauth.associations.find()
print(json.dumps(association_list))

# Output: 
# [{"app_guid": "9dfa33ee-62ea-4a69-8b1a-51735c70dba9", "oauth_integration_guid": "9dd7ef9d-ccd4-46f0-b76c-12b00f1ff594", "oauth_integration_name": "default-test", "oauth_integration_description": "desc", "oauth_integration_template": "custom", "created_time": "2024-09-12T19:05:11Z"}]

# integration_b also sees the association
integration = c.oauth.integrations.get(integration_b_guid)
association_list = integration.associations.find()
print(json.dumps(association_list))

# Output: 
# [{"app_guid": "9dfa33ee-62ea-4a69-8b1a-51735c70dba9", "oauth_integration_guid": "9dd7ef9d-ccd4-46f0-b76c-12b00f1ff594", "oauth_integration_name": "default-test", "oauth_integration_description": "desc", "oauth_integration_template": "custom", "created_time": "2024-09-12T19:05:11Z"}]

# integration_a does not see the integration
integration = c.oauth.integrations.get(integration_a_guid)
association_list = integration.associations.find()
print(json.dumps(association_list))

# Output: 
# []

```

src/posit/connect/content.py
```diff
@@ -27,6 +41,12 @@ def __getitem__(self, key: Any) -> Any:
             return ContentItemOwner(params=self.params, **v)
         return v
 
+    @property
+    def oauth(self) -> ContentItemOAuth:
+        if "guid" not in self:
+            raise ValueError("ContentItemOAuth requires content guid")
```
tdstein, line 47: I think you can remove this. If 'guid' is missing, we have bigger issues. And I believe the default KeyError thrown by self['guid'] if 'guid' is missing is sufficient.

src/posit/connect/oauth/integrations.py
```diff
@@ -2,12 +2,24 @@
...
 
 class Integration(Resource):
     """OAuth integration resource."""
 
+    @property
+    def associations(self) -> IntegrationAssociations:
+        if "guid" not in self:
+            raise ValueError(
+                "IntegrationAssociations requires integration guid"
+            )
```
tdstein, line 18: Same as above.

tdstein (approved):
This looks great! I have a few minor nits on exception handling, but I'm ok with merging the pull request as is or with changes.

src/posit/connect/content.py
zackverham: cool, yep - makes sense. I was getting type checking errors when we were using property-based fields because the guids were `str|None` types, but the change to a `self['guid']` lookup removes that issue entirely. 

I'll make that change and then merge when CI passes.

PR #288, opened 2024-09-12
build: use pyright as static type checker

tdstein:
When analyzing overload definitions, Pyright does a better job at understanding the conventions we have established. This is due to how Pyright has decided to interpret PEP-484. Additional details are available here: https://github.com/microsoft/pyright/blob/main/docs/type-concepts-advanced.md#overloads

Additionally, Pyright discovered a few issues with types, which are addressed in this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ it:
 	$(MAKE) -C ./integration
 
 lint:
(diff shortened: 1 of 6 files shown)


zackverham (approved):
LGTM

PR #289, opened 2024-09-13
build: change default to latest

tdstein:
This changes the behavior of the `make it` command from "./Makefile".

The `make it` command is only used for convenience. Otherwise, the "./integration/Makefile" is used directly for integration testing.


tdstein:
@zackverham, you may find this useful for your current workstream.

toph-allen (approved):

PR #290, opened 2024-09-13
fix: content overload signatures

### LINKED ISSUE #140: Better docstrings for content.update

tdstein:
Simplifies or, in most cases, corrects the overload signature. Variable names and documentation are updated along the way.

- Renames `kwargs` based on context. The variable name `conditions` is introduced for querying methods. The variable name `attributes` is introduced for record manipulation. This also signifies that this is not a passthrough method. 
- Documentation is updated to reflect the current state of Posit Connect.
- Removes deprecations for field names in integration tests.

Resolves #140

### DIFF
--- a/integration/tests/posit/connect/test_content.py
+++ b/integration/tests/posit/connect/test_content.py
@@ -12,7 +12,7 @@ class TestContent:
     @classmethod
     def setup_class(cls):
         cls.client = connect.Client()
-        cls.content = cls.client.content.create(name="example")
+        cls.content = cls.client.content.create()
 
     @classmethod
     def teardown_class(cls):
@@ -23,7 +23,7 @@ class TestContent:
         assert self.client.content.count() == 1
 
     def test_get(self):
-        assert self.client.content.get(self.content.guid) == self.content
+        assert self.client.content.get(self.content["guid"]) == self.content
 
     def test_find(self):
         assert self.client.content.find()
@@ -33,13 +33,15 @@ class TestContent:
 
     def test_content_item_owner(self):
         item = self.client.content.find_one(include=None)
+        assert item
         owner = item.owner
-        assert owner.guid == self.client.me.guid
+        assert owner["guid"] == self.client.me["guid"]
 
     def test_content_item_owner_from_include(self):
         item = self.client.content.find_one(include="owner")
+        assert item
         owner = item.owner
-        assert owner.guid == self.client.me.guid
+        assert owner["guid"] == self.client.me["guid"]
 
     @pytest.mark.skipif(
         CONNECT_VERSION <= version.parse("2024.04.1"),
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -444,7 +444,10 @@ class Content(Resources):
         *,
         name: Optional[str] = None,
         owner_guid: Optional[str] = None,
-        include: str | list[Literal["owner", "tags", "vanity_url"]] = [],
+        include: Optional[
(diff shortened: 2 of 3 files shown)


integration/tests/posit/connect/test_content.py
```diff
@@ -23,7 +23,7 @@ def test_count(self):
         assert self.client.content.count() == 1
 
     def test_get(self):
-        assert self.client.content.get(self.content.guid) == self.content
```
dbkegley, line 26: I'm reaching the edge of my python knowledge here so maybe this is a dumb question but is `.guid` now a deprecated syntax and did that happen in this PR?

dbkegley (approved):

integration/tests/posit/connect/test_content.py
tdstein: It's self-imposed. See https://github.com/posit-dev/posit-sdk-py/issues/262

We are moving away from attribute-based access for object data.

src/posit/connect/content.py
```diff
@@ -119,76 +119,108 @@ def restart(self) -> None:
...
+    def update(self, **attributes: Any) -> None:
+        """Update the content."""
         ...
 
-    def update(self, *args, **kwargs) -> None:
-        """Update the content item."""
-        body = dict(*args, **kwargs)
-        url = self.params.url + f"v1/content/{self.guid}"
-        response = self.params.session.patch(url, json=body)
+    def update(self, **attributes: Any) -> None:
```
zackverham, line 220: ahh, I love this - definitely much easier to understand what's going on.

tdstein: 🙌🏻

src/posit/connect/content.py
```diff
@@ -119,76 +119,108 @@ def restart(self) -> None:
...
+        """Update the content."""
+        url = self.params.url + f"v1/content/{self['guid']}"
```
zackverham, line 222: not sure how to weigh pros / cons here, just an observation - when I'm interacting with an SDK resource, my LSP would give me autocomplete on the deprecated property fields, which I don't get using the new approach. Its harder to explore what fields are available on the resource, or in some cases know where to look at all for that information if it isn't in a docstring.  

Its also easier for the LSP to catch typos on properties - if I were to typo a dict key, I would get a more confusing error.

zackverham (approved):

src/posit/connect/content.py
tdstein: Good callout. That is a big downside to this approach and the main reason we tried using attributes originally. I have experimented with defining a TypedDict and then holding data in an internal attribute instead of on self. I abandoned it due to the complexity of maintaining the facade between the internal state and self. I'm open to revisiting it, though.

The other answer is... you always have that problem when using dictionaries in Python. So this isn't anything new  🙃

zackverham: yeah, it might just be a matter of making sure we're good with our documentation of resources so we can easily check what's in them.

PR #291, opened 2024-09-13
style: set line length to 99

tdstein:
The maximum line length recommended in PEP-8 is 99.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ version_file = "src/posit/_version.py"
 
 [tool.ruff]
 extend-exclude = ["integration/resources"]
+line-length = 99
 
 [tool.ruff.format]
 docstring-code-format = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,6 @@ version_file = "src/posit/_version.py"
 
 [tool.ruff]
 extend-exclude = ["integration/resources"]
-line-length = 79
 
 [tool.ruff.format]
 docstring-code-format = true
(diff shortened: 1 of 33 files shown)


tdstein:
https://peps.python.org/pep-0008/#maximum-line-length

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

I prefer a longer line length. For what it's worth, the GitHub diff view side-by-side fits comfortably on my screen at this length.

pyproject.toml
```diff
@@ -37,7 +37,7 @@ version_file = "src/posit/_version.py"
 
 [tool.ruff]
 extend-exclude = ["integration/resources"]
-line-length = 79
+line-length = 99
```
dbkegley, line 40: How did you decide on 99? I need to update neovim with a config like this so that I get a visual when my lines are too long (mostly for doc writing since markdown handles wrapping automatically) but I can't decide on a good global default. The standard terminal width 80 feels too small and I think that would be annoying

dbkegley (approved):

pyproject.toml
tdstein: Does NeoVim support line wraps?

dbkegley: Probably but I don't actually want the automatic wrapping in most cases because it tends to cause such large diffs when I edit files in repos that don't have any formatting enforcement. I prefer the visual so I can see when I'm editing a line that exceeds the length and I can just fix it while I'm in there for lines that I'm already changing.

PR #296, opened 2024-09-18
feat: add find_by method to content

tdstein:
Adds the find_by method to content. The find_by method is a convenience method for filtering a collection of records by attribute values.  All filtering happens client-side. This method can be improved to utilize server-side filtering using query parameters.

### DIFF
--- a/integration/tests/posit/connect/test_content.py
+++ b/integration/tests/posit/connect/test_content.py
@@ -29,7 +29,7 @@ class TestContent:
         assert self.client.content.find()
 
     def test_find_by(self):
-        assert self.client.content.find_by(guild=self.content["guid"]) == self.content
+        assert self.client.content.find_by(guid=self.content["guid"]) == self.content
 
     def test_find_one(self):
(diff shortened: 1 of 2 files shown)


tests/posit/connect/test_content.py
```diff
@@ -406,6 +406,45 @@ def test_params_include_none(self):
...
+        mock_get = responses.get(
+            "https://connect.example/__api__/v1/content",
+            json=load_mock("v1/content.json"),
+        )
+
+        # setup
+        client = Client("https://connect.example", "12345")
+
+        # invoke
+        content = client.content.find_by(name="does-not-exist")
```
zackverham, line 441: if I tried passing in an invalid keyword arg through the **attributes parameter, would anything bad happen?

zackverham (approved):

tests/posit/connect/test_content.py
tdstein: Nothing bad. You just won't get any results.

PR #299, opened 2024-09-26
feat: add version check support

### LINKED ISSUE #272: Add the Connect version as a resource attribute.
It's happening! We're adding support for an API endpoint that only exists on newer Connect versions (#267).

We need to raise a `NotImplementError` when calling these methods against a Connect version older than August 2024. One way to approach this is by checking the result of `client.version`. Let's add some core logic that injects this information into all resources.

Important! By default, Connect returns this information in the server settings API. But, this functionality can be turned off via Connect configuration. Therefore, we cannot always depend on this method for obtaining the Connect version. We need to accommodate these scenarios. A simple approach assumes the server supports the API if the Connect version is unknown. Then, use a requests hook to fail in a sensible way.

tdstein:
Adds a new decorator, `@context.requires`, which asserts version compatibility when the server version is known. The check is skipped if the server version is unknown (e.g., the Connect configuration disables version information).

Also marks the OAuth API with a '2024.08.0' requirement.

Closes #272

### DIFF
--- a/src/posit/connect/context.py
+++ b/src/posit/connect/context.py
@@ -1,13 +1,15 @@
 import functools
 from typing import Optional, Protocol
 
+from packaging.version import Version
+
 
 def requires(version: str):
     def decorator(func):
         @functools.wraps(func)
         def wrapper(instance: ContextManager, *args, **kwargs):
             ctx = instance.ctx
-            if ctx.version and ctx.version < version:
+            if ctx.version and Version(ctx.version) < Version(version):
                 raise RuntimeError(
                     f"This API is not available in Connect version {ctx.version}. Please upgrade to version {version} or later.",
                 )
--- a/src/posit/connect/context.py
+++ b/src/posit/connect/context.py
@@ -1,0 +1,43 @@
+import functools
+from typing import Optional, Protocol
+
+
+def requires(version: str):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(instance: ContextManager, *args, **kwargs):
+            ctx = instance.ctx
+            if ctx.version and ctx.version < version:
+                raise RuntimeError(
+                    f"This API is not available in Connect version {ctx.version}. Please upgrade to version {version} or later.",
+                )
+            return func(instance, *args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+class Context(dict):
+    def __init__(self, session, url):
+        self.session = session
+        self.url = url
+
+    @property
+    def version(self) -> Optional[str]:
+        try:
+            value = self["version"]
+        except KeyError:
+            endpoint = self.url + "server_settings"
+            response = self.session.get(endpoint)
+            result = response.json()
+            value = self["version"] = result.get("version")
+        return value
+
+    @version.setter
+    def version(self, value: str):
+        self["version"] = value
+
+
+class ContextManager(Protocol):
+    ctx: Context
--- a/tests/posit/connect/external/test_databricks.py
+++ b/tests/posit/connect/external/test_databricks.py
@@ -48,6 +48,7 @@ class TestPositCredentialsHelpers:
         register_mocks()
 
         client = Client(api_key="12345", url="https://connect.example/")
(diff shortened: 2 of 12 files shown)


src/posit/connect/context.py
```diff
@@ -0,0 +1,43 @@
+import functools
+from typing import Optional, Protocol
+
+
+def requires(version: str):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(instance: ContextManager, *args, **kwargs):
+            ctx = instance.ctx
+            if ctx.version and ctx.version < version:
```
tdstein, line 10: Need to update this with better version comparison using `packaging.version.Version` or something else.

tests/posit/connect/external/test_databricks.py
```diff
@@ -48,6 +48,7 @@ def test_posit_credentials_provider(self):
         register_mocks()
 
         client = Client(api_key="12345", url="https://connect.example/")
+        client.ctx.version = "2024.08.0"
```
tdstein, line 51: This shortcuts having to mock the API request to /server_settings.

tests/posit/connect/external/test_databricks.py
```diff
@@ -48,6 +48,7 @@ def test_posit_credentials_provider(self):
         register_mocks()
 
         client = Client(api_key="12345", url="https://connect.example/")
+        client.ctx.version = None
```
tdstein, line 51: We are setting the version to `None` to avoid mocking the API request to /server_settings.

tdstein:
After a few iterations, I landed on the decorator approach. This seems idiomatic and self-documenting. I tried a few declarative approaches (i.e., checking the version inline), but that felt brittle.

tests/posit/connect/oauth/test_oauth.py
```diff
@@ -23,5 +23,6 @@ def test_get_credentials(self):
                 "token_type": "Bearer",
             },
         )
-        con = Client(api_key="12345", url="https://connect.example/")
-        assert con.oauth.get_credentials("cit")["access_token"] == "viewer-token"
+        c = Client(api_key="12345", url="https://connect.example/")
+        c.ctx.version = None
```
dbkegley, line 27: Should we test here that the oauth client resource respects the server's version constraint?

dbkegley (approved):
Looks great! I like the annotation approach, that feels very idiomatic

tests/posit/connect/external/test_databricks.py
jonkeane: Should we put this in a helper function of some sort? Something like `test_client(...)` that does this internally. There are other ways too, but this is a lot of the same thing over and over (and a bit of specialized knowledge that you basically _always_ have to do this in these tests

tests/posit/connect/test_context.py
```diff
@@ -0,0 +1,90 @@
...
+
+            @requires("1.0.0")
+            def success(self):
+                pass
+
+        ctx = MagicMock()
+        ctx.version = None
+        instance = Stub(ctx)
+
+        instance.success()
```
jonkeane, line 57: This test is for behavior we use while testing, but not one we expect users to hit, yeah? If yeah, it would be good to comment about that just so it's clear this isn't a desired end-user behavior.

tests/posit/connect/oauth/test_integrations.py
```diff
@@ -164,10 +167,11 @@ def test(self):
         )
 
         # setup
-        client = Client("https://connect.example", "12345")
+        c = Client("https://connect.example", "12345")
+        c.ctx.version = None
```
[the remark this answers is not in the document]
tdstein: Just making the references the same across the file.

tests/posit/connect/test_context.py
tdstein: No, this is designed behavior. If the version is hidden in the server settings, then `ctx.version` will be None. When the value is `None` the version check is skipped since we aren't sure. In the case of breaking API changes, we will need to perform the fallback sequence based on the HTTP response.

PR #300, opened 2024-09-26
feat: Content repository resource

zackverham:
@tdstein looking for feedback here on the semantics / feel of this implementation. 

If I have a test script `scratch.py`: 

```
from posit.connect.client import Client

c = Client("http://localhost:3939", "<redacted>")

valid_content_guid="89c71a26-9dee-4b5a-b3af-cf135ef6079a"
# get content item
content_item = c.content.get(valid_content_guid)
# get the content item's repo
repo = content_item.repository()
print(repo)
# update something about the repo
repo.update(branch="this-is-an-updated-branch")
print(repo)
# delete the repo
repo.delete()
# error currently
print(content_item.repository())
```

here's what the result looks like:
```
(posit-sdk-py) ➜  posit-sdk-py git:(zack-192-git-repo-settings) ✗ python scratch.py
{'repository': 'https://github.com/dbkegley/rsc-sample-content', 'branch': 'main', 'directory': 'local/python-shiny', 'polling': True}
{'repository': 'https://github.com/dbkegley/rsc-sample-content', 'branch': 'this-is-an-updated-branch', 'directory': 'local/python-shiny', 'polling': True}
Traceback (most recent call last):
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/scratch.py", line 16, in <module>
    print(content_item.repository())
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/src/posit/connect/content.py", line 160, in repository
    response = self.params.session.get(url)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/.venv/lib/python3.12/site-packages/requests/sessions.py", line 602, in get
    return self.request("GET", url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/.venv/lib/python3.12/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/.venv/lib/python3.12/site-packages/requests/sessions.py", line 710, in send
    r = dispatch_hook("response", hooks, r, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/.venv/lib/python3.12/site-packages/requests/hooks.py", line 30, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zverham/Development/posit-dev/posit-sdk-py/src/posit/connect/hooks.py", line 18, in handle_errors
    raise ClientError(error_code, message, http_status, http_status_message, payload)
posit.connect.errors.ClientError: {"error_code": 4, "error_message": "The requested object does not exist.", "http_status": 404, "http_message": "Not Found", "payload": null}
```

Current things I'm not sure about: 

- The use of `repository()` as a function to get the repo information. Because we don't have a guid associated with the git info, the pattern `content.repository.get()` can't overload the base `get` function correctly, so I'm not sure the best way to express a singleton `get` operation that will never be parameterized. 

- If no repository exists, how do we want to represent that? The API itself returns an API error - I don't know how we want to represent this situation in the SDK.


tdstein:
Nice!

I've been encountering this problem as well. Since the Repository is one-to-one with the Content, the interface is different, and the existing Resource/Resources base classes do not work well with this concept.

Strictly thinking about the public interface, I agree that using `repository()` is weird. Instead, I would prefer to call `print(content.repository)` and see the response body of the GET repository endpoint. If the Repository doesn't exist, I would like to see that somehow.

Things get tricky when you introduce the concept of creating a Repository. I think `content.repository.create(...)` feels natural. It turns out you can solve this problem using classmethods. Here's an example:

```py

repositories = []


class Repository(dict):
    @classmethod
    def create(cls, **attributes):
        # simulated POST endpoint
        repository = Repository(**attributes)
        repositories.append(repository)
        return repository


class Content(dict):
    @property
    def repository(self):
        # simulated GET endpoint
        return next(iter(repositories), Repository)


if __name__ == "__main__":
    content = Content()
    # Nothing here...
    print(content.repository)

    # Let's create one!
    repository = content.repository.create(url="git+https://github.com/posit-dev/posit-sdk-py")
    print(repository)

    # There it is!
    print(content.repository)
```

There are still some oddities here. For example, the first call to `content.repository` prints `<class '__main__.Repository'>`. Ideally, we make this a little more intuitive.

zackverham:
New and improved ergonomics: 

```
from posit.connect.client import Client

c = Client("http://localhost:3939", "<redacted>")

valid_content_guid="9086fef1-a600-4eb0-afbd-51791df32efa"
invalid_content_guid = "9dfa33ee-62ea-4a69-8b1a-51735c70dba9"

# get content item
content_item = c.content.get(valid_content_guid)
repo = content_item.repository
if repo:
    repo.update(branch="this-is-an-updated-branch")
    print(repo)

    repo.delete()
    print(content_item.repository)
else:
    print("oops, repo does not exist")
```

src/posit/connect/_api.py
```diff
@@ -0,0 +1,261 @@
...
+class ApiContextProtocol(Protocol):
+    _ctx: Context
+    _path: str
+
+
+# TODO-future?; Add type hints for the ReadOnlyDict class
+# ArgsT = TypeVar("ArgsT", bound="ResponseAttrs")
+
+
+class ReadOnlyDict(
```
tdstein, line 33: For what it's worth, I think this is fairly similar to the original functionality implemented and then removed in this commit: https://github.com/posit-dev/posit-sdk-py/commit/50e78ac5b8d608c34144553db9727203045a04bb

Locking the __setitem__ post initialization is interesting. Did you explore using `frozendict`? https://pypi.org/project/frozendict/

It's worth considering if __setitem__ should make a PATCH request to have the server reflect the local object state. Based on our current conversations, I am leaning towards no. Instead, a call to `update` should be used. 

Hm... If we modified `Resource` to inherit from `frozendict` instead of `dict`, then a lot of the `update` behavior is simplified.

I pulled on the thread a little bit... Here is a modification to `Resource` to use `frozendict`. It may be odd to look at, but I think it's worth exploring further.

```py
class Resource(frozendict):
    params: ResourceParameters

    def __new__(cls: Type["Resource"], params: ResourceParameters, *args, **kwargs):
        instance = super(Resource, cls).__new__(cls, *args, **kwargs)
        instance.params = params
        return instance

    def __init__(self, params: ResourceParameters, *args, **kwargs):
        # Skip setting attributes in __init__ to maintain immutability
        super().__init__(*args, **kwargs)

    def __getattr__(self, name):
        if name in self:
            warnings.warn(
                f"Accessing the field '{name}' via attribute is deprecated and will be removed in v1.0.0. "
                f"Please use __getitem__ (e.g., {self.__class__.__name__.lower()}['{name}']) for field access instead.",
                DeprecationWarning,
                stacklevel=2,
            )
            return self[name]
        return None
        ```

**Review by tdstein — COMMENTED** [review]
Checking in to see how the Git Repository implementation is going. Looks like you have some exciting things in progress. Feel free to disregard any of my comments if they aren't relevant to where you are headed.

If making Resource immutable or revisiting attributes for fields is worth pursuing, we should split that work out from this pull request to make sure the core Git Repository functionality makes this release.

File: src/posit/connect/_api.py (hunk 1 above)
Comment (schloerke) [reply src/posit/connect/_api.py:33]: I have not looked through `frozendict` before. Neat project!
Nice trick with `__new__`. Clever.

-----------------------------

Couple ideas I was trying to go for in `ApiDictEndpoint` which inherits from `EndpointMixin` and this `ReadOnlyDict`.

1. Do not expose `.params`. It is internal information. (Redistributed to `._ctx` and `._path`)
2. Do not expose any methods through implementation coincidence. Ex: `frozendict` is awesome but exposes methods like `.setdefault(key, default)`, which I do not want to show to users.

Maybe a [`collections.abc.Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) would work? It would only exposes `keys`, `items`, `values`, and `get` (plus `__getitem__`, `__iter__`, `__len__`, `__contains__`, `__eq__`, and `__ne__`).

Comment (schloerke) [reply src/posit/connect/_api.py:33]: Using a read only dictionary makes it quite clear!

File: pyproject.toml [hunk 2]
```diff
@@ -70,6 +70,7 @@ extend-select = [
     # Check docstring formatting. Many of these rules are intentionally ignored below.
     "D",
 
+    "ARG",    # ARG; flake8-argparse: https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg
```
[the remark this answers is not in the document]
schloerke: Correct. Faster turn around with Ruff

src/posit/connect/content.py
```diff
@@ -140,16 +294,15 @@ def restart(self) -> None:
...
         self,
         *,
         # Required argument
-        name: str,
+        name: Optional[str] = None,
         # Content Metadata
         title: Optional[str] = None,
         description: Optional[str] = None,
-        access_type: Literal["all", "acl", "logged_in"] = "acl",
+        access_type: Optional[Literal["all", "acl", "logged_in"]] = None,
```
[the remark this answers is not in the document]
schloerke: Same as above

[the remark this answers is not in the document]
schloerke: The difficult part is that the default value represents an actual value. 

When updating, only the values being set should be updated. If there is a default value, then that value should be used unless overwritten.

It was never an issue because the implementation previously did not _actually_ use the default value... It was just an `@override`

src/posit/connect/content.py
```diff
@@ -234,15 +388,36 @@ def update(
...
+            "amd_gpu_limit": amd_gpu_limit,
+            "nvidia_gpu_limit": nvidia_gpu_limit,
+            "run_as": run_as,
+            "run_as_current_user": run_as_current_user,
+            "default_image_name": default_image_name,
+            "default_r_environment_management": default_r_environment_management,
+            "default_py_environment_management": default_py_environment_management,
+            "service_account_name": service_account_name,
+            **attributes,
+        }
```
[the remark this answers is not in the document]
schloerke: Yes!

I believe the `ContentItemRepository` has the proper implementation

src/posit/connect/resources.py
```diff
@@ -26,19 +32,26 @@ class ResourceParameters:
...
         if name in self:
-            warnings.warn(
-                f"Accessing the field '{name}' via attribute is deprecated and will be removed in v1.0.0. "
-                f"Please use __getitem__ (e.g., {self.__class__.__name__.lower()}['{name}']) for field access instead.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
+            # TODO-future; Temp silence during all testing; Turn it into an error.
+            global has_shown_warning_dict
+            if not has_shown_warning_dict:
```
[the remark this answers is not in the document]
schloerke: Correct. It was super annoying when testing. So I print the first and turn the rest off.

We should remove it before merging

src/posit/connect/resources.py
```diff
@@ -201,7 +214,7 @@ def find(self, uid) -> T:
         result = response.json()
         return self._to_instance(result)
 
-    def find_by(self, **conditions: Any) -> Optional[T]:
+    def find_by(self, **conditions: Any) -> T | None:
```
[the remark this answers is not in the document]
schloerke: I believe it'll work for py3.8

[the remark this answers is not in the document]
tdstein: Last time I tried I was seeing a TypeError: https://github.com/posit-dev/posit-sdk-py/pull/315#discussion_r1824674190

Let me know if you find a way around it. I agree that the `T | None` syntax is ideal.

src/posit/connect/content.py
```diff
@@ -53,6 +174,49 @@ def __getitem__(self, key: Any) -> Any:
...
+
+        Returns
+        -------
+        ContentItemRepository
+        """
+        response = ContentItemRepository(
+            self._ctx,
+            content_guid=self["guid"],
+            # Add a placeholder `attr` to avoid a `GET` request.
+            _init=True,  # pyright: ignore[reportCallIssue]
```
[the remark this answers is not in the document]
schloerke: Yes. Agreed. Moved logic into the class

src/posit/connect/resources.py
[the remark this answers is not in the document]
schloerke: By requiring the use of 

```
```

We can use `|` syntax on _not top level_ usage. 

```
Bad = int | str

class Ok:
    def method(self) -> int | str: ...
```

(This check is now made within ruff and will give a good suggested fix 😄 )

schloerke:
Comment: 
> Look into using the `Client` class within `ApiDictEndpoint` to make api calls.

The functionality does not overlap as `Client` wants to return `requests.Response`, where as `ApiCallMixin` wants to call `.json()` on the response. The methods are small enough that the logic can be duplicated without the need to pass around the root object.

src/posit/connect/_api.py
schloerke: Talked through w/ @tdstein in person and am using `Mapping` with minimal method exposure as our way forward.

src/posit/connect/content.py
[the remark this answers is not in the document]
schloerke: When using `._Attrs` unpacking, there are no defaults... skirting the issue.

Closing

src/posit/connect/content.py
```diff
@@ -359,11 +533,12 @@ def create(
         nvidia_gpu_limit: Optional[int] = None,
         # Execution Settings
         run_as: Optional[str] = None,
-        run_as_current_user: Optional[bool] = False,
+        run_as_current_user: Optional[bool] = None,
```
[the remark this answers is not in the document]
schloerke: When using `._Attrs` unpacking, there are no defaults... skirting the issue.

Closing

src/posit/connect/content.py
```diff
@@ -331,16 +506,15 @@ def count(self) -> int:
         """
         return len(self.find())
 
-    @overload
     def create(
         self,
         *,
         # Required argument
-        name: str,
+        name: Optional[str] = None,
```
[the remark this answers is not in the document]
schloerke: `._Attrs` requires name for all the methods using it.

Closing

src/posit/connect/content.py
```diff
@@ -331,16 +506,15 @@ def count(self) -> int:
...
         # Content Metadata
         title: Optional[str] = None,
         description: Optional[str] = None,
-        access_type: Literal["all", "acl", "logged_in"] = "acl",
+        access_type: Optional[Literal["all", "acl", "logged_in"]] = None,
```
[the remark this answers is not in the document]
schloerke: When using `._Attrs` unpacking, there are no defaults... skirting the issue.

Closing

src/posit/connect/content.py
```diff
@@ -169,11 +322,12 @@ def update(
         nvidia_gpu_limit: Optional[int] = None,
         # Execution Settings
         run_as: Optional[str] = None,
-        run_as_current_user: Optional[bool] = False,
+        run_as_current_user: Optional[bool] = None,
```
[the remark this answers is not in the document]
schloerke: When using `._Attrs` unpacking, there are no defaults... skirting the issue.

Closing

src/posit/connect/_api.py
```diff
@@ -0,0 +1,278 @@
...
+
+        Returns
+        -------
+        T
+            The first record matching the conditions, or `None` if no match is found.
+        """
+        results = self.fetch()
+        for item in results:
+            if obj_has_attrs(item, conditions):
+                return item
```
tdstein, line 267: https://github.com/posit-dev/posit-sdk-py/blob/1dacc4dd1e7560af2080d887328e7f1d4dc8b1de/src/posit/connect/resources.py#L220

src/posit/connect/_api.py
```diff
@@ -0,0 +1,278 @@
...
+        uid : str
+            The unique identifier of the record.
+
+        Returns
+        -------
+        :
+            Single instance of T if found, else None
+        """
+        return self.find_by(**{self._uid_key: uid})
+
```
tdstein, line 248: https://github.com/posit-dev/posit-sdk-py/blob/1dacc4dd1e7560af2080d887328e7f1d4dc8b1de/src/posit/connect/resources.py#L202

src/posit/connect/_api.py
```diff
@@ -0,0 +1,278 @@
...
+        return self._attrs.get(key, default)
+
+    def __getitem__(self, key: str) -> Any:
+        return self._attrs[key]
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        raise AttributeError(
+            "Resource attributes are locked. "
+            "To retrieve updated values, please retrieve the parent object again."
+        )
```
tdstein, line 69: I believe this should be a `NotImplementedError` instead of an `AttributeError`.

src/posit/connect/_api.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def obj_has_attrs(obj: ReadOnlyDict, attrs: dict[str, Any]) -> bool:
+    for attr_key, attr_val in attrs.items():
+        if not hasattr(obj, attr_key):
+            return False
+        obj_val = getattr(obj, attr_key)
+        if obj_val != attr_val:
+            return False
+    return True
```
tdstein, line 278: todo - remove

src/posit/connect/_api_call.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _endpoint(self, extra_endpoint: str = "") -> str: ...
+    def _get_api(self, *, extra_endpoint: str = "") -> Jsonifiable: ...
+    def _delete_api(self, *, extra_endpoint: str = "") -> Jsonifiable | None: ...
+    def _patch_api(self, json: Jsonifiable | None, *, extra_endpoint: str = "") -> Jsonifiable: ...
+    def _put_api(self, json: Jsonifiable | None, *, extra_endpoint: str = "") -> Jsonifiable: ...
+
+
+def endpoint(ctx: Context, path: str, *, extra_endpoint: str = "") -> str:
+    return ctx.url + path + extra_endpoint
```
tdstein, line 22: pathinfo is the conventional name for this. You could also use `*args`. You should use posixpath.join to build the path to ensure all parts are separated by '/'

Something like...
```suggestion
def endpoint(ctx: Context, *path) -> str:
    return ctx.url + posixpath.join(*path)
```

src/posit/connect/_api_call.py
```diff
@@ -0,0 +1,65 @@
...
+
+class ApiCallProtocol(Protocol):
+    _ctx: Context
+    _path: str
+
+    def _endpoint(self, extra_endpoint: str = "") -> str: ...
+    def _get_api(self, *, extra_endpoint: str = "") -> Jsonifiable: ...
+    def _delete_api(self, *, extra_endpoint: str = "") -> Jsonifiable | None: ...
+    def _patch_api(self, json: Jsonifiable | None, *, extra_endpoint: str = "") -> Jsonifiable: ...
+    def _put_api(self, json: Jsonifiable | None, *, extra_endpoint: str = "") -> Jsonifiable: ...
```
tdstein, line 18: I think it would be wise to lean on the Python requests Session interface. Which does runtime assertions to verify the types. Our testing strategy will reveal any issues related to the incorrect type being used.

src/posit/connect/_typing_extensions.py
tdstein, line 1: This is nice. I wasn't aware of the issues in typing_extensions..

src/posit/connect/hooks.py
```diff
@@ -6,7 +6,7 @@
 from .errors import ClientError
 
 
-def handle_errors(response: Response, *args, **kwargs) -> Response:
+def handle_errors(response: Response, *args, **kwargs) -> Response:  # noqa: ARG001
```
tdstein, line 9: Is it possible to disable "ARG001" globally for *args and **kwargs only?

src/posit/connect/resources.py
```diff
@@ -26,19 +31,27 @@ class ResourceParameters:
...
+            if not has_shown_warning_dict:
+                # # Uncomment for local testing
+                # has_shown_warning_dict = True
+                warnings.warn(
+                    f"Accessing the field '{name}' via attribute is deprecated and will be removed in v1.0.0. "
+                    f"Please use __getitem__ (e.g., {self.__class__.__name__.lower()}['{name}']) for field access instead.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
             return self[name]
```
tdstein, line 55: Would you mind removing the has_shown_warning_dict. It's fairly simple to suppress these errors. And it's a good nudge to resolve the deprecation during implementation.

tests/posit/connect/test_content.py
```diff
@@ -542,3 +547,84 @@ def test_app_mode_is_other(self):
...
+            assert new_repository_info[key] == value
+
+    @responses.activate
+    def test_repository_delete(self):
+        repository_info = self.mock_repository_info()
+
+        mock_delete = responses.delete(self.endpoint)
+        repository_info.delete()
+
+        assert mock_delete.call_count == 1
```
tdstein, line 630: Ideally, these tests should follow the existing pattern used in the other test. The json should be stored in a file in the __api__ folder that represents a mocked server response.

integration/tests/posit/connect/test_content_item_repository.py
```diff
@@ -0,0 +1,94 @@
...
+        assert updated_repo["repository"] == self.repo_repository
+        assert updated_repo["directory"] == self.repo_directory
+        assert updated_repo["polling"] is self.repo_polling
+
+        # Delete
+        content.repository.delete()
+        assert content.repository is None
+
+        # Cleanup
+        content.delete()
```
tdstein, line 94: This should happen in teardown test since other classes use the same "example" name for content. Otherwise, if this fails, those tests will throw an error.

src/posit/connect/hooks.py
schloerke: I do not see a way.

Can I just remove the arguments?

src/posit/connect/_typing_extensions.py
schloerke: 💙

src/posit/connect/_api_call.py
schloerke: Perfect!

src/posit/connect/_api.py
schloerke: 👍

tdstein:
@schloerke - I went through and removed all __getattr__ calls which triggered the warning in `Resource`: https://github.com/posit-dev/posit-sdk-py/pull/323

src/posit/connect/_api_call.py
schloerke: Moved to issue

tdstein (changes requested):
Functionality looks good. Just need few minor edits for consistency, then we can merge 👍🏻

src/posit/connect/content.py
```diff
@@ -22,25 +33,190 @@
...
+
+        result = put_api(ctx, cls._api_path(content_guid), json=cast(JsonifiableDict, attrs))
+
+        return ContentItemRepository(
+            ctx,
+            content_guid=content_guid,
+            **result,  # pyright: ignore[reportCallIssue]
+        )
+
+    def delete(self) -> None:
```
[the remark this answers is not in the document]
tdstein: I have been moving to destroy to logically separate from a common `__del__` operation

src/posit/connect/_api.py
```diff
@@ -0,0 +1,285 @@
...
+        super().__init__(attrs)
+        self._ctx = ctx
+        self._path = path
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+    def __repr__(self) -> str:
+        # CLASSNAME - v1/content/123/path; {'guid': '123', 'name': 'My Content'}
+        return repr(f"{self.__class__.__name__} - {self._path}; {super().__repr__()}")
```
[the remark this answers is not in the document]
schloerke: I'd like to leave it in as the information is useful.

Will address it again on merging of resource class.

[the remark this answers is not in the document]
tdstein: No, the following expression must be true, which it is not: `eval(repr(repository)) == repository`.

Please use the default methods from `dict` for the time being. In the future, we can look into implementing true `__repr__` which would provide true serliaization functionality if it is necessary. See https://docs.python.org/3/library/functions.html#repr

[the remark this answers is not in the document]
schloerke: I'll remove it.

tdstein (approved):

PR #305, opened 2024-10-08
feat: add create local user support

### LINKED ISSUE #113: Users.create()
This will probably break down into multiple issues given that there are several ways to create users.

### LINKED ISSUE #225: Verify `client.user.content` works when multiple users exist.
Depends on #113 

Add additional integration tests to `integration/tests/posit/connect/test_users.py` that check behavior when multiple users exist.

tdstein:
Adds create user support for local authentication providers (e.g., 'ldap', 'oauth2', 'pam', 'password', 'proxy', or 'saml').

Closes #113
Closes #225

### DIFF
--- a/integration/tests/posit/connect/test_users.py
+++ b/integration/tests/posit/connect/test_users.py
@@ -1,14 +1,55 @@
 from posit import connect
 
 
-class TestAttributeContent:
+class TestUser:
+    @classmethod
+    def setup_class(cls):
+        cls.client = client = connect.Client()
+        cls.aron = client.users.create(
+            username="aron", email="aron@example.com", password="s3cur3p@ssword"
+        )
+        cls.bill = client.users.create(
+            username="bill", email="bill@example.com", password="s3cur3p@ssword"
+        )
+        cls.cole = client.users.create(
+            username="cole", email="cole@example.com", password="s3cur3p@ssword"
+        )
+
+    def test_lock(self):
+        for user in (self.aron, self.bill, self.cole):
+            user.lock()
+            assert len(self.client.users.find(account_status="locked")) == 1
+            user.unlock()
+        assert len(self.client.users.find(account_status="locked")) == 0
+
+    def test_count(self):
+        # aron, bill, cole, and me
+        assert self.client.users.count() == 4
+
+    def test_find(self):
+        assert self.client.users.find(prefix="aron") == [self.aron]
+        assert self.client.users.find(prefix="bill") == [self.bill]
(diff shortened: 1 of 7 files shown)


integration/tests/posit/connect/test_users.py
```diff
@@ -1,29 +1,79 @@
...
+        cls.client = client = connect.Client()
+        cls.aron = client.users.create(
+            username="aron", email="aron@example.com", password="s3cur3p@ssword"
+        )
+        cls.bill = client.users.create(
+            username="bill", email="bill@example.com", password="s3cur3p@ssword"
+        )
+        cls.cole = client.users.create(
+            username="cole", email="cole@example.com", password="s3cur3p@ssword"
+        )
```
toph-allen, line 16: For my own understanding: why are three users being tested? Are they different in some way?

src/posit/connect/users.py
```diff
@@ -117,21 +159,188 @@ class Users(Resources):
...
+
+        Returns
+        -------
+        User
+            The newly created user.
+        """
+        # todo - use the 'context' module to inspect the 'authentication' object and route to POST (local) or PUT (remote).
+        url = self.params.url + "v1/users"
+        response = self.params.session.post(url, json=attributes)
+        return User(self.params, **response.json())
```
toph-allen, line 268: Remind me: what's the purpose of each successive `@overload`? I'm sure it's written out somewhere.

toph-allen: I get that some of the overloads are documentation — it just is non-obvious to me reading this how the two I've highlighted here interact.

toph-allen (approved):
All of my questions are just "for my own understanding" type things.

integration/tests/posit/connect/test_users.py
tdstein: They probably aren't strictly necessary, but I wanted to ensure the filtering worked correctly.

src/posit/connect/users.py
tdstein: "How they interact" may be the wrong way to think about it. The overload definitions "[are for the benefit of the type checker only](https://docs.python.org/3/library/typing.html#typing.overload)." The final overload definition says you can call this method within any keyword arguments (`**attributes`). 

I've chosen this to say that you can call this method with any arguments, and they will be passed as the request body. This covers cases where some versions of Connect accept a request body that is different from those defined in the previous overload method definitions. Ideally, we have correct overload definitions for every version of Connect. But, this works as a catch-all fallback in cases where we are unaware.

Would it be helpful to include this information as part of the docstring?

PR #310, opened 2024-10-09
feat: add vanities

### LINKED ISSUE #175: Implement Content Vanity API
# Content Vanity API

## v0.6.0 Proposal

```python
from posit import connect

client = connect.Client()
content = client.content.get(...)

vanity = content.vanity = ...
del content.vanity

vanity.destroy()
```

tdstein:
Adds vanities support.

A new mixin pattern is introduced, which adds the `vanity` attributes to the `ContentItem` class. This abstraction is still a bit leaky since the endpoint is hardcoded to 'v1/content' . But I find this design helpful since it collocates everything regarding vanities into a single file. In addition, the Content class has started to get a little unwieldy due to its size.

The use of TypedDict, Unpack, Required, and NotRequired from `typing_extensions` is also introduced. Thanks to https://blog.changs.co.uk/typeddicts-are-better-than-you-think.html, I discovered this pattern provides a much more succinct way to define pass-through arguments. It also greatly simplifies the verbosity of overload methods. That isn't shown here, but I have a test on another branch.

Resolves #175


tdstein:
Sorry for the churn on this. While writing out the cookbook recipes, I discovered a few improvements. This should be good to go now.

tdstein:
Ok, I'm pretty sure I'm actually done now :)

src/posit/connect/vanities.py
```diff
@@ -0,0 +1,236 @@
...
+        after_destroy: Optional[AfterDestroyCallback] = None,
+        **kwargs: Unpack[VanityAttributes],
+    ):
+        """Initialize a Vanity.
+
+        Parameters
+        ----------
+        params : ResourceParameters
+        after_destroy : AfterDestroyCallback, optional
+            Called after the Vanity is successfully destroyed, by default None
```
toph-allen, line 68: This is interesting to me; what are its use cases?

src/posit/connect/vanities.py
```diff
@@ -0,0 +1,236 @@
...
+
+class VanityMixin(Resource):
+    """Mixin class to add a vanity attribute to a resource."""
+
+    class HasGuid(TypedDict):
+        """Has a guid."""
+
+        guid: Required[str]
+
+    def __init__(self, /, params: ResourceParameters, **kwargs: Unpack[HasGuid]):
```
toph-allen, line 128: What does `/` mean?

src/posit/connect/vanities.py
```diff
@@ -0,0 +1,236 @@
...
+        return [Vanity(self.params, **result) for result in results]
+
+
+class VanityMixin(Resource):
+    """Mixin class to add a vanity attribute to a resource."""
+
+    class HasGuid(TypedDict):
+        """Has a guid."""
+
+        guid: Required[str]
```
toph-allen, line 126: Could you explain how this works? Is the internal class definition here is mostly for type-related stuff? (I see it used in the `__init__` method below)

tests/posit/connect/test_vanities.py
```diff
@@ -0,0 +1,116 @@
...
+    @responses.activate
+    def test_destroy_calls_after_destroy_callback(self):
+        content_guid = "8ce6eaca-60af-4c2f-93a0-f5f3cddf5ee5"
+        base_url = "http://connect.example/__api__"
+        endpoint = f"{base_url}/v1/content/{content_guid}/vanity"
+        responses.delete(endpoint)
+
+        session = requests.Session()
+        url = Url(base_url)
+        after_destroy = Mock()
```
toph-allen, line 38: Aha — is `after_destroy` a parameter so that it can be mocked?

toph-allen (approved):

src/posit/connect/vanities.py
tdstein: It's currently used on L145 below (`self._vanity._after_destroy = self.reset_vanity`), which says, "after the vanity is destroyed, call reset_vanity". It's a way to control the state of VanityMixin without having to interact with it directly from Vanity.

src/posit/connect/vanities.py
tdstein: Yep, it is strictly for the type checker. It makes this class easier to work with in unit tests.

src/posit/connect/vanities.py
tdstein: https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters

Although, I no longer need this. It's cruft left over from implementation.

```suggestion
    def __init__(self, params: ResourceParameters, **kwargs: Unpack[HasGuid]):
```

PR #312, opened 2024-10-15
ci: add requests compatiblity check

tdstein:


.github/workflows/deps.yaml
```diff
@@ -0,0 +1,100 @@
...
+          - '2.4.3'
+          - '2.4.2'
+          - '2.4.1'
+          - '2.4.0'
+          - '2.3.0'
+          - '2.2.1'
+          - '2.2.0'
+          - '2.1.0'
+          - '2.0.1'
+          - '2.0.0'
```
nealrichardson, line 89: This feels like overkill. Why not just test on 2.0.0, and if we find that we need to add a specific version later, we can?

tdstein: You're probably right. This is just an attempt at understanding where things break.  This latest commit is broken, but if you look [here](https://github.com/posit-dev/posit-sdk-py/actions/runs/11348910107), you will see that things break starting with 2.27.1 and earlier. This issue stems from how we are catching errors in the hooks module.

tdstein:
Progress stalled. Closing for now.

PR #314, opened 2024-10-23
feat: add jobs

### LINKED ISSUE #306: Add Content Jobs endpoints

tdstein:
Adds job support.

The mixin pattern is continued, and a 'jobs' property is added to `ContentItem.`

A new abstraction around resources called `Active` is introduced. This abstraction begins a transition to a more idiomatic approach to interacting with collections. Similar to how printing a `Resource` shows a `dict`, the `ActiveSequence` shows a familiar `List[dict]` representation. The `ActiveFinderMethods` supports the `find` and `find_by` methods. This is inspired by this Ruby on Rails module https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html. Ideally, this separation of concerns provides a straightforward way to compose bindings in the future and reduces the amount of duplication across implementations.

Additional background may be found here https://github.com/tdstein/active

Related to #310 
Resolves #306

### DIFF
--- a/integration/tests/posit/connect/test_jobs.py
+++ b/integration/tests/posit/connect/test_jobs.py
@@ -1,0 +1,29 @@
+from pathlib import Path
+
+from posit import connect
+
+
+class TestContent:
+    @classmethod
+    def setup_class(cls):
+        cls.client = connect.Client()
+        cls.content = cls.client.content.create(name="example-quarto-minimal")
+
+    @classmethod
+    def teardown_class(cls):
+        cls.content.delete()
+        assert cls.client.content.count() == 0
+
+    def test(self):
+        content = self.content
+
+        path = Path("../../../resources/connect/bundles/example-quarto-minimal/bundle.tar.gz")
+        path = Path(__file__).parent / path
+        path = path.resolve()
+        path = str(path)
+
+        bundle = content.bundles.create(path)
+        bundle.deploy()
+
+        jobs = content.jobs
+        assert len(jobs) == 1
--- a/integration/tests/posit/connect/test_jobs.py
+++ b/integration/tests/posit/connect/test_jobs.py
@@ -1,6 +1,11 @@
 from pathlib import Path
 
+import pytest
+from packaging import version
+
 from posit import connect
+
+from . import CONNECT_VERSION
 
 
 class TestContent:
@@ -14,6 +19,10 @@ class TestContent:
         cls.content.delete()
         assert cls.client.content.count() == 0
 
+    @pytest.mark.skipif(
+        CONNECT_VERSION <= version.parse("2023.01.1"),
+        reason="Quarto not available",
+    )
     def test(self):
         content = self.content
 
--- a/integration/tests/posit/connect/test_jobs.py
+++ b/integration/tests/posit/connect/test_jobs.py
@@ -8,7 +8,7 @@ from posit import connect
 from . import CONNECT_VERSION
 
 
-class TestContent:
+class TestJobs:
     @classmethod
     def setup_class(cls):
         cls.client = connect.Client()
--- a/src/posit/connect/jobs.py
+++ b/src/posit/connect/jobs.py
@@ -1,20 +1,132 @@
-from typing import List, Sequence, TypedDict
+from typing import List, Literal, Optional, Sequence, TypedDict, overload
 
-from typing_extensions import Required, Unpack
+from typing_extensions import NotRequired, Required, Unpack
 
-from .resources import Resource, ResourceParameters, Resources
+from .errors import ClientError
+from .resources import FinderMethods, Resource, ResourceParameters, Resources
+
+JobTag = Literal[
+    "unknown",
+    "build_report",
+    "build_site",
+    "build_jupyter",
+    "packrat_restore",
+    "python_restore",
+    "configure_report",
+    "run_app",
+    "run_api",
+    "run_tensorflow",
+    "run_python_api",
+    "run_dash_app",
+    "run_streamlit",
+    "run_bokeh_app",
+    "run_fastapi_app",
+    "run_pyshiny_app",
+    "render_shiny",
+    "run_voila_app",
+    "testing",
+    "git",
+    "val_py_ext_pkg",
+    "val_r_ext_pkg",
+    "val_r_install",
+]
 
 
 class Job(Resource):
-    pass
+    class _Job(TypedDict):
+        # Identifiers
+        id: Required[str]
+        """A unique identifier for the job."""
+
+        ppid: Required[Optional[str]]
+        """Identifier of the parent process."""
+
+        pid: Required[str]
+        """Identifier of the process running the job."""
+
+        key: Required[str]
+        """A unique key to identify this job."""
+
+        remote_id: Required[Optional[str]]
+        """Identifier for off-host execution configurations."""
+
+        app_id: Required[str]
+        """Identifier of the parent content associated with the job."""
+
+        variant_id: Required[str]
+        """Identifier of the variant responsible for the job."""
+
+        bundle_id: Required[str]
+        """Identifier of the content bundle linked to the job."""
+
+        # Timestamps
+        start_time: Required[str]
+        """RFC3339 timestamp indicating when the job started."""
+
+        end_time: Required[Optional[str]]
+        """RFC3339 timestamp indicating when the job finished."""
+
+        last_heartbeat_time: Required[str]
+        """RFC3339 timestamp of the last recorded activity for the job."""
+
+        queued_time: Required[Optional[str]]
+        """RFC3339 timestamp when the job was added to the queue."""
+
+        # Status and Exit Information
+        status: Required[Literal[0, 1, 2]]
+        """Current status. Options are 0 (Active), 1 (Finished), and 2 (Finalized)"""
+
+        exit_code: Required[Optional[int]]
+        """The job's exit code, available after completion."""
+
+        # Environment Information
+        hostname: Required[str]
+        """Name of the node processing the job."""
+
+        cluster: Required[Optional[str]]
+        """Location where the job runs, either 'Local' or the cluster name."""
+
+        image: Required[Optional[str]]
+        """Location of the content in clustered environments."""
+
+        run_as: Required[str]
+        """UNIX user responsible for executing the job."""
+
+        # Queue and Scheduling Information
+        queue_name: Required[Optional[str]]
+        """Name of the queue processing the job, relevant for scheduled reports."""
+
+        # Job Metadata
+        tag: Required[JobTag]
+        """A tag categorizing the job type. Options are build_jupyter, build_report, build_site, configure_report, git, packrat_restore, python_restore, render_shiny, run_api, run_app, run_bokeh_app, run_dash_app, run_fastapi_app, run_pyshiny_app, run_python_api, run_streamlit, run_tensorflow, run_voila_app, testing, unknown, val_py_ext_pkg, val_r_ext_pkg, and val_r_install."""
+
+    def __init__(self, /, params, endpoint, **kwargs: Unpack[_Job]):
+        super().__init__(params, **kwargs)
+        key = kwargs["key"]
+        self._endpoint = endpoint + key
+
+    def destroy(self) -> None:
+        """Destroy the job.
+
+        Submit a request to kill the job.
+
+        Warnings
+        --------
+        This operation is irreversible.
+
+        Note
+        ----
+        This action requires administrator, owner, or collaborator privileges.
+        """
+        self.params.session.delete(self._endpoint)
 
 
-class Jobs(Resources, Sequence[Job]):
+class Jobs(FinderMethods[Job], Sequence[Job], Resources):
     """A collection of jobs."""
 
     def __init__(self, params, endpoint):
-        super().__init__(params)
-        self._endpoint = endpoint
+        super().__init__(Job, params, endpoint)
+        self._endpoint = endpoint + "jobs"
         self._cache = None
 
     @property
@@ -24,7 +136,7 @@ class Jobs(Resources, Sequence[Job]):
 
         response = self.params.session.get(self._endpoint)
         results = response.json()
-        self._cache = [Job(self.params, **result) for result in results]
+        self._cache = [Job(self.params, self._endpoint, **result) for result in results]
         return self._cache
 
     def __getitem__(self, index):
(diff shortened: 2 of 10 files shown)


src/posit/connect/jobs.py
```diff
@@ -0,0 +1,273 @@
...
+        """Identifier of the variant responsible for the job."""
+
+        bundle_id: NotRequired[str]
+        """Identifier of the content bundle linked to the job."""
+
+        # Timestamps
+        start_time: NotRequired[str]
+        """RFC3339 timestamp indicating when the job started."""
+
+        end_time: NotRequired[Optional[str]]
```
toph-allen, line 169: What's the difference between `NotRequired` and `Optional`?

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+            Additional keyword arguments passed to the parent `Resource` class.
+        """
+        params = ResourceParameters(ctx.session, ctx.url)
+        super().__init__(params, **kwargs)
+        self._ctx = ctx
+        self._parent = parent
+
+
+T_co = TypeVar("T_co", bound="Active", covariant=True)
+"""A covariant type variable that is bound to the `Active` class, meaning that `T_co` must be or derive from `Active`."""
```
toph-allen, line 94: What is the purpose of this restriction?

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+            If no record is found.
+        """
+        if self._cache:
+            conditions = {self._uid: uid}
+            result = self.find_by(**conditions)
+        else:
+            endpoint = self._endpoint + uid
+            response = self._ctx.session.get(endpoint)
+            result = response.json()
+            result = self._cls(self._ctx, self._parent, **result)
```
toph-allen, line 277: Could use talking through how the operators work here &… well, I feel fuzzy on a few questions about how this works when you're finding by something that isn't the key. Does it get all the jobs from the endpoint and find within those? I know I'm misunderstanding some part of this relationship.

tests/posit/connect/test_jobs.py
```diff
@@ -0,0 +1,157 @@
...
+    def test(self):
+        responses.get(
+            "https://connect.example/__api__/v1/content/f2f37341-e21d-3d80-c698-a935ad614066",
+            json=load_mock("v1/content/f2f37341-e21d-3d80-c698-a935ad614066.json"),
+        )
+
+        responses.get(
+            "https://connect.example/__api__/v1/content/f2f37341-e21d-3d80-c698-a935ad614066/jobs",
+            json=load_mock("v1/content/f2f37341-e21d-3d80-c698-a935ad614066/jobs.json"),
+        )
```
toph-allen, line 99: In cases like this, it'd be helpful to see comments explaining what each of the different mocks are testing (if they test different conditions), since that information isn't available here.

toph-allen (approved):
This generally looks good — but approving with the caveat that I don't feel like I fully grok some of the mechanisms at play here. Would be great to walk through the PR together so I can ask questions.

src/posit/connect/resources.py
tdstein, line 94:
```suggestion
"""A covariant type variable that is bound to the `Active` class and must inherit from it."""
```

integration/tests/posit/connect/test_jobs.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @pytest.mark.skipif(
+        CONNECT_VERSION <= version.parse("2023.01.1"),
+        reason="Quarto not available",
+    )
+    def test(self):
+        content = self.content
+
+        path = Path("../../../resources/connect/bundles/example-quarto-minimal/bundle.tar.gz")
+        path = Path(__file__).parent / path
```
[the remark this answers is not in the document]
schloerke: When using `Path` from `pathlib`, the `/` operator will concatenate the two values together, returning a `Path` object.

`__file__` represents the file of this code: integration/tests/posit/connect/test_jobs.py
The end result of `path` will be `` plus `"../../../resources/connect/bundles/example-quarto-minimal/bundle.tar.gz"`, so `"integration/resources/connect/bundles/example-quarto-minimal/bundle.tar.gz" ([link](https://github.com/posit-dev/posit-sdk-py/tree/9e4f17348da963cdf4f777f7ae0c4fc8b6434428/integration/resources/connect/bundles/example-quarto-minimal))

src/posit/connect/jobs.py
```diff
@@ -0,0 +1,273 @@
...
+        tag: Required[JobTag]
+        """A tag categorizing the job type. Options are build_jupyter, build_report, build_site, configure_report, git, packrat_restore, python_restore, render_shiny, run_api, run_app, run_bokeh_app, run_dash_app, run_fastapi_app, run_pyshiny_app, run_python_api, run_streamlit, run_tensorflow, run_voila_app, testing, unknown, val_py_ext_pkg, val_r_ext_pkg, and val_r_install."""
+
+    def __init__(self, ctx, parent: Active, **kwargs: Unpack[_Job]):
+        super().__init__(ctx, parent, **kwargs)
+        self._parent = parent
+
+    @property
+    def _endpoint(self) -> str:
+        return self._ctx.url + f"v1/content/{self._parent['guid']}/jobs/{self['key']}"
```
schloerke, line 108: Can the `._ctx`, `._parent['guid']`, or `self['key']` change over time?
Why not use a direct property instead? (Set within the `__init__` method)

src/posit/connect/jobs.py
schloerke: `Optional[FOO]` represents `FOO | None`. 

I'm in the camp of function parameters can be defined using `Optional` (as it hints that the parameter is _optional_), but everything else should be defined as `FOO | None`.

In this case, `NotRequired` is a special "this key does not have to exist within the dictionary". So `end_time` does not have to exist within a `_FindByRequest` object. (But if the key does exist, it can be a `str` or `None`)

src/posit/connect/jobs.py
```diff
@@ -0,0 +1,273 @@
...
+        queue_name: NotRequired[Optional[str]]
+        """Name of the queue processing the job, relevant for scheduled reports."""
+
+        # Job Metadata
+        tag: NotRequired[JobTag]
+        """A tag categorizing the job type. Options are build_jupyter, build_report, build_site, configure_report, git, packrat_restore, python_restore, render_shiny, run_api, run_app, run_bokeh_app, run_dash_app, run_fastapi_app, run_pyshiny_app, run_python_api, run_streamlit, run_tensorflow, run_voila_app, testing, unknown, val_py_ext_pkg, val_r_ext_pkg, and val_r_install."""
+
+    @overload
+    def find_by(self, **conditions: Unpack[_FindByRequest]) -> Optional[Job]:
+        """Finds the first record matching the specified conditions.
```
schloerke, line 208: Docs go in the base method?

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+    def __len__(self) -> int:
+        return len(self._data)
+
+    def __str__(self) -> str:
+        return str(self._data)
+
+    def __repr__(self) -> str:
+        return repr(self._data)
+
+    def reload(self) -> "ActiveSequence":
```
schloerke, line 213: What about `typing.Self` to return the "self" object to maintain any generic values. (It isn't a _new_ ActiveSequence. It is the same one.)

```suggestion
    def reload(self) -> Self:
```

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+        be queried to fetch the data.
+
+        Returns
+        -------
+        str
+            The API endpoint URL.
+        """
+        raise NotImplementedError()
+
+    @property
```
schloerke, line 173: Because this is doing more than a special retrieval, I'd argue for a true method over a property... Let's pkg authors know it's doing something and not just immediately returning a value

schloerke: By having a method, it also removes the possibility of setting the value

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+        Return the number of items in the sequence.
+
+    __str__() -> str
+        Return a string representation of the cached data.
+
+    __repr__() -> str
+        Return a detailed string representation of the cached data.
+
+    reload() -> ActiveSequence
+        Clear the cache and mark to reload the data from the API on the next operation.
```
schloerke, line 147: This is fairly tedious to maintain. 

I hope we can transition to [quartodoc](https://github.com/machow/quartodoc/).  It will pull all the docs and fill this out for you, avoiding the painful double record keeping of types and docs.

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+    -------
+    find(uid) -> T_co
+        Finds and returns a record by its unique identifier (`uid`). If a cached result exists, it searches within the cache;
+        otherwise, it makes a GET request to retrieve the data from the endpoint.
+
+    find_by(**conditions: Any) -> Optional[T_co]
+        Finds the first record that matches the provided conditions. If no record is found, returns None.
+    """
+
+    _uid: str = "guid"
```
schloerke, line 247: Shiny has been burned by using default values within classes. By having it in the class definition, the value is bound to the class, not really to the instance. Finding the root cause of the bug wasn't fun.  To be safe, init all class values within the `__init__` method.

If this can be moved to within the init function, then the situation is avoided.

src/posit/connect/jobs.py
```diff
@@ -0,0 +1,273 @@
...
+    def find_by(self, **conditions): ...
+
+    def find_by(self, **conditions) -> Optional[Job]:
+        return super().find_by(**conditions)
+
+
+class JobsMixin(Active, Resource):
+    """Mixin class to add a jobs attribute to a resource."""
+
+    def __init__(self, ctx, **kwargs):
```
schloerke, line 271: This feels like an abuse of inheritance to achieve execution. I'd argue the `Jobs(Job, ctx, self)` should be put into a function.

```python
def make_jobs(ctx: Context, parent: Active):
    return Jobs(Job, ctx, parent)
```

integration/tests/posit/connect/test_jobs.py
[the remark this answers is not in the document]
tdstein: That's right. Thanks @schloerke!

src/posit/connect/jobs.py
tdstein: `NotRequired` indicates that the argument may be supplied or not supplied. If it is not supplied, no default value is set.

`Optional`, on the other hand, means that the value can be either None or the provided type.

For this example, `NotRequired[Optional[str]]` means that you have three options as the caller.
1. `find_by(end_time="2023-05-25T00:10:01-02:00")` - Find any match with this exact timestamp.
2. `find_by(end_time=None)` - Find any match where the end_time is None (`null` in JSON).
3. `find_by()` - Find any match (using other arguments)

tdstein:
@toph-allen and I met over Zoom to discuss his questions and talk about the code in more depth. Please let me know if anyone else would like to do the same. Happy to do so.

src/posit/connect/resources.py
tdstein: Yes, it is indeed. I generally use an LLM to speed up the process.

We have Quartodoc setup (see the `docs` folder), but I haven't been consistent with maintaining it. If there is any easier way, please advise! 😄

src/posit/connect/resources.py
tdstein: I can see that. Since it is bound to the class, I'll see if __init_subclass__ can work.

src/posit/connect/jobs.py
tdstein: Typically, yes, an attribute would be right. This is only a property method because it is an abstractmethod / property in the base class.

src/posit/connect/jobs.py
tdstein: This is only for the Language Server Protocol. If you start typing a call to `find_by` in VSCode / Positron, you'll be prompted with the docstring for each overload method independently.

src/posit/connect/resources.py
```diff
@@ -43,3 +46,254 @@ def update(self, *args, **kwargs):
...
+        Returns
+        -------
+        ActiveSequence
+            The current instance with cleared cache, ready to reload data on next access.
+        """
+        self._cache = None
+        return self
+
+
+class ActiveFinderMethods(ActiveSequence[T_co], ABC, Generic[T_co]):
```
schloerke, line 226: I believe this can be dropped as `ActiveSequence` covers it. (untested)

```suggestion
class ActiveFinderMethods(ActiveSequence[T_co], ABC):
```

src/posit/connect/resources.py
schloerke: I believe `__init_subclass__` also works on the class object. 

I don't readily see any string method that allows for altering in place. So it might be ok as is. 

If it was a more complex / mutable object object, it would run into problems

src/posit/connect/resources.py
tdstein: After a discussion with @schloerke, we realized this doesn't need to be covariant. I've updated the usage to `TypeVar("T", bound="Active")`

src/posit/connect/jobs.py
```diff
@@ -0,0 +1,292 @@
...
+    "val_py_ext_pkg",
+    "val_r_ext_pkg",
+    "val_r_install",
+]
+
+
+class Job(Active):
+    class _Job(TypedDict):
+        # Identifiers
+        id: Required[str]
```
zackverham, line 37: I like getting type definitions back!

src/posit/connect/resources.py
```diff
@@ -43,3 +47,196 @@ def update(self, *args, **kwargs):
 class Resources:
     def __init__(self, params: ResourceParameters) -> None:
         self.params = params
+
+
+class Active(ABC, Resource):
+    def __init__(self, ctx: Context, parent: Optional["Active"] = None, **kwargs):
+        """A base class representing an active resource.
```
zackverham, line 54: a dumb question - what does "active" mean in this context? Are we naming it that as a callback to the Rails reference implementation? Or is this meaning something connect-specific?

tdstein: Yeah, it's just a callback to the Rails name. Open to suggestions if anyone has a better name idea.

src/posit/connect/jobs.py
```diff
@@ -0,0 +1,292 @@
+from typing import Literal, Optional, TypedDict, overload
+
+from typing_extensions import NotRequired, Required, Unpack
+
+from .resources import Active, ActiveFinderMethods, ActiveSequence, Resource
+
+JobTag = Literal[
```
zackverham, line 7: a nit beyond all nits: wondering if this list should be organized a bit more cleanly - we could alphabetize within job types or something like that. 

if Connect adds new job types and starts returning them to the SDK, will that cause issues?

tdstein: It wouldn't break anything. The type system would be outdated, but we expect this to happen over time. Since we provide wildcard fallbacks for public methods (e.g., find_by, create, etc.), the user's program shouldn't be affected.

src/posit/connect/resources.py
```diff
@@ -223,31 +171,25 @@ def reload(self) -> "ActiveSequence":
...
 
-class ActiveFinderMethods(ActiveSequence[T_co], ABC, Generic[T_co]):
-    """
-    Finder methods.
+class ActiveFinderMethods(ActiveSequence[T], ABC, Generic[T]):
+    def __init__(self, ctx: Context, parent: Optional[Active] = None, uid: str = "guid"):
+        """Finder methods.
 
-    Provides various finder methods for locating records in any endpoint supporting HTTP GET requests.
+        Provides various finder methods for locating records in any endpoint supporting HTTP GET requests.
```
schloerke, line 178: When we get to quartodoc updates, we will most likely need to update where these docs are located. But it should just be shuffling them around and chopping type definitions.

src/posit/connect/resources.py
tdstein: > Does it get all the jobs from the endpoint and find within those?

That's correct.

PR #315, opened 2024-10-25
refactor: inject url path parts instead of endpoints

tdstein:
Refactors path building responsibilities to the creating action, eliminating a ton of complexity along the way.


src/posit/connect/vanities.py
```diff
@@ -231,6 +225,7 @@ def find_vanity(self) -> Vanity:
         -------
         Vanity
         """
-        response = self.params.session.get(self._endpoint)
+        endpoint = self.params.url + f"v1/content/{self._content_guid}/vanity"
```
toph-allen, line 228: It kinda feels weird to me to see this change from a computed property to a repeated format string 🤔

src/posit/connect/jobs.py
```diff
@@ -120,40 +122,42 @@ def destroy(self) -> None:
...
         ----------
         ctx : Context
-            The context containing the HTTP session used to interact with the API.
-        parent : Active
-            Parent resource for maintaining hierarchical relationships
+            The context object containing the session and URL for API interactions
+        path : str
+            The HTTP path component for the collection endpoint
+        pathinfo : str
+            The HTTP part of the path directed at a specific resource, by default "jobs"
```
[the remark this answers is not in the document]
toph-allen: Or is the `path` collection here the content item and `path info` the resource hanging off that content item? In that case, it doesn't really make sense to have `pathinfo` as as a parameter you can redefine here?

toph-allen (commented):
I think I need a bit more of an explanation about how this is working before I fully understand it.

src/posit/connect/vanities.py
tdstein: Makes sense. Realistically, this is a temporary change since this module will be rewritten to use the Active bases.

src/posit/connect/jobs.py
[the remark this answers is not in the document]
tdstein: > Like, this makes me wonder what the point of being able to redefine the jobs part of the URL for the Jobs collection class is.

You're right, it could be removed here since we don't have variation in the endpoint path between Connect versions.

> is the path collection here the content item and path info the resource hanging off that content item?

I may be simpler to look at this within the `ActiveSequence` class. The reason it is split into path and pathinfo is due to how the paths are built. If you trace the creation of the `Job` (singular) `_path` variable, you'll see that it is derived from the `Content` it hangs off of, plus the path parts "jobs" and the `Job` guid.

src/posit/connect/jobs.py
```diff
@@ -99,13 +100,14 @@ class _Job(TypedDict):
...
         """A tag categorizing the job type. Options are build_jupyter, build_report, build_site, configure_report, git, packrat_restore, python_restore, render_shiny, run_api, run_app, run_bokeh_app, run_dash_app, run_fastapi_app, run_pyshiny_app, run_python_api, run_streamlit, run_tensorflow, run_voila_app, testing, unknown, val_py_ext_pkg, val_r_ext_pkg, and val_r_install."""
 
-    def __init__(self, ctx, parent: Active, **kwargs: Unpack[_Job]):
-        super().__init__(ctx, parent, **kwargs)
-        self._parent = parent
+    @overload
+    def __init__(self, ctx: Context, path: str, pathinfo: str, /, **attributes: Unpack[_Job]): ...
+
+    @overload
+    def __init__(self, ctx: Context, path: str, pathinfo: str, /, **attributes: Any): ...
```
schloerke, line 107: Its too bad these overloads are necessary

src/posit/connect/jobs.py
```diff
@@ -120,40 +122,42 @@ def destroy(self) -> None:
...
+            The field name used to uniquely identify records, by default "guid"
         """
-        super().__init__(ctx, parent, uid)
-        self._parent = parent
+        super().__init__(ctx, path, pathinfo, uid)
 
-    @property
-    def _endpoint(self) -> str:
-        return self._ctx.url + f"v1/content/{self._parent['guid']}/jobs"
+    def _create_instance(self, path: str, pathinfo: str, **kwargs: Any) -> Job:
```
schloerke, line 146: Add in arg separator

```suggestion
    def _create_instance(self, path: str, pathinfo: str, /, **kwargs: Any) -> Job:
```

src/posit/connect/resources.py
```diff
@@ -171,29 +176,12 @@ def reload(self) -> Self:
...
-            The context containing the HTTP session used to interact with the API.
-        parent : Optional[Active], optional
-            Optional parent resource for maintaining hierarchical relationships, by default None
-        uid : str, optional
-            The default field name used to uniquely identify records, by default "guid"
-        """
-        super().__init__(ctx, parent)
-        self._uid = uid
-
+class ActiveFinderMethods(ActiveSequence[T], ABC):
```
schloerke, line 179: Restoring as class docs

```suggestion
class ActiveFinderMethods(ActiveSequence[T], ABC):
    """Finder methods.

    Provides various finder methods for locating records in any endpoint supporting HTTP GET requests.""""
```

schloerke: Yes, this does stomp the docs added above, but maybe in a week or two we can do a quartodoc overhual session. This will make the boiler plate disappear and documentation could go with its the type definition

src/posit/connect/resources.py
```diff
@@ -125,9 +123,16 @@ def _data(self) -> List[T]:
         if self._cache:
```
schloerke, line 123: I believe empty lists are also false here. 

We should probably compare against `self._cache is not None`

src/posit/connect/resources.py
```diff
@@ -203,24 +191,23 @@ def find(self, uid) -> T:
...
+                return result
+
+        endpoint = self._ctx.url + self._path + uid
+        response = self._ctx.session.get(endpoint)
+        result = response.json()
+        result = self._create_instance(self._path, uid, **result)
+
+        # Invalidate the cache.
+        # It is assumed that the cache is stale since a record exists on the server and not in the cache.
+        self._cache = None
```
schloerke, line 210: Why not set directly to `result` as its already an result of `._create_instance()`?

src/posit/connect/vanities.py
```diff
@@ -71,10 +71,6 @@ def __init__(
         self._after_destroy = after_destroy
         self._content_guid = kwargs["content_guid"]
 
-    @property
-    def _endpoint(self):
-        return self.params.url + f"v1/content/{self._content_guid}/vanity"
-
```
schloerke, line 77: Just clarifying my comment from earlier...

I'm not against the helper function `def _endpoint(self):` , I just don't want it as a `@property`.

src/posit/connect/resources.py
```diff
@@ -50,64 +51,61 @@ def __init__(self, params: ResourceParameters) -> None:
...
-        be queried to fetch the data.
-
-        Returns
-        -------
-        str
-            The API endpoint URL.
-        """
-        raise NotImplementedError()
-
     @property
```
schloerke, line 109: This method performs a lot of side effects. I would suggest dropping `@property` to help signal things are being performed.

```suggestion
```

schloerke (changes requested):
- [ ] Drop `@property`
- [ ] Discuss cache invalidation of `self._cache`

src/posit/connect/jobs.py
tdstein: They've grown on me. A large part of this project is documentation. Even more so now that we are abstracting out the API calls.

src/posit/connect/resources.py
tdstein: `_cache` holds the collection, while `result` is a single instance in this case.

src/posit/connect/vanities.py
tdstein: Ah, I see the distinction you are making. Thanks for the clarification. Either way, this will go away when migrating this module to use `Active`.

src/posit/connect/resources.py
tdstein: Sounds great!

src/posit/connect/resources.py
tdstein: Good call! It turns out I've been misusing the property annotation 🙃

src/posit/connect/jobs.py
schloerke: From out conversation...

Fallback overload was added to avoid breaking types when users use new API values that are not implemented.

While it works, it breaks type checking for all other variables. I'm in the stance of "If the user chose to add the new parameter. Then they can ignore the type checking for the parameter."

So I would ask to remove the fallback overload.

src/posit/connect/jobs.py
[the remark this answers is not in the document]
schloerke: I really like how `parent` was removed in favor of the url information!

---------------

At the very bottom of the usage stack, both variables are joined together and not leveraged anywhere else. 

Given Toph's suggestion, could we
* Drop `path` from `Job.__init__()` parameter and up through to `JobsMixin.__init__()`
	* This could set a standard of only passing required information
	* Allows the class to hide away the `path` information as it is constant
Or
* Make `path` within `Content` and pass the single value through. (Essentially dropping all `path_info` parameters.
	* Low mental overhead, but deviates from the `path` wiki definition.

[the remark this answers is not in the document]
tdstein: I removed the path and pathinfo arguments when they were unnecessary. See https://github.com/posit-dev/posit-sdk-py/pull/315/commits/03accf84423a70c53db7b41373ad5ebda7b85414

I think this makes the most sense in practice.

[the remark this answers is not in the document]
tdstein: Eh, spoke too soon. Doing a bit more cleanup now...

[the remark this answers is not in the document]
tdstein: Ok done.

tdstein:
@schloerke  / @toph-allen—This is ready for final review. I believe all of the points we discussed have been addressed.

I am pretty happy with the implementation now. Thanks for all of your thoughts and attention to detail!

src/posit/connect/resources.py
```diff
@@ -203,26 +204,17 @@ def find(self, uid) -> T:
...
-        if not result:
-            raise ValueError(f"Failed to find instance where {self._uid} is '{uid}'")
+            if result:
+                return result
 
-        return result
+        endpoint = self._ctx.url + self._path + uid
+        response = self._ctx.session.get(endpoint)
+        result = response.json()
+        return self._to_instance(result)
```
tdstein, line 217: Note that I removed the call to invalidate the cache that existed a few commits before. After some additional consideration, I concluded that invalidating the cache is an unwanted side effect.

I think there is still an argument for invalidating the cache or appending the instance to the cached list. But, I don't think we have a good enough understanding of the side effects to proceed with either implementation.

src/posit/connect/resources.py
```diff
@@ -61,77 +61,108 @@ def __init__(self, ctx: Context, path: str, pathinfo: str = "", /, **attributes)
...
 """A type variable that is bound to the `Active` class"""
 
 
 class ActiveSequence(ABC, Generic[T], Sequence[T]):
-    def __init__(self, ctx: Context, path: str, pathinfo: str = "", uid: str = "guid"):
-        """A sequence abstraction for any HTTP GET endpoint that returns a collection.
+    """A sequence for any HTTP GET endpoint that returns a collection."""
 
-        It lazily fetches data on demand, caches the results, and allows for standard sequence operations like indexing and slicing.
+    _cache: Optional[List[T]]
```
schloerke, line 81:
```suggestion
    _cache: List[T] | None
```

src/posit/connect/resources.py
```diff
@@ -61,77 +61,108 @@ def __init__(self, ctx: Context, path: str, pathinfo: str = "", /, **attributes)
...
         self._uid = uid
-        self._cache: Optional[List[T]] = None
+        self._cache = None
+
+    @abstractmethod
+    def _create_instance(self, path: str, /, **kwargs: Any) -> T:
+        """Create an instance of 'T'."""
+        raise NotImplementedError()
+
+    def cached(self) -> bool:
```
schloerke, line 106: More descriptive name / make private as its more for internal use than subclass use.

```suggestion
    def _has_cache(self) -> bool:
```

src/posit/connect/resources.py
```diff
@@ -61,77 +61,108 @@ def __init__(self, ctx: Context, path: str, pathinfo: str = "", /, **attributes)
...
+        reload
         """
-        Fetch and cache the data from the API.
+        return self._cache is not None
 
-        This method sends a GET request to the `_endpoint` and parses the response as a list of JSON objects.
-        Each JSON object is used to instantiate an item of type `T` using the class specified by `_cls`.
-        The results are cached after the first request and reused for subsequent access unless reloaded.
+    def reload(self) -> Self:
+        """Reloads the collection from Connect.
```
schloerke, line 120: Mark for internal usage. Match name to actions performed
```suggestion
    def _reset_cache(self) -> Self:
        """Resets the collected values from Connect.
```

src/posit/connect/resources.py
```diff
@@ -61,77 +61,108 @@ def __init__(self, ctx: Context, path: str, pathinfo: str = "", /, **attributes)
...
+        Returns
+        -------
+        List[T]
 
+        See Also
+        --------
+        cached
+        reload
+        """
+        if self._cache is None:
```
schloerke, line 164:
```suggestion
        if not self._has_cache():
```

schloerke (approved):

src/posit/connect/resources.py
schloerke: Why not just call `.find_by()` all the time? 

If the cache exists, `._data` returns quickly. If not, it asks the server.

Then both methods have the same quirks. (Where as `find()` will not alter the cache, but find_by will... causing a followup call to `find()` to use the cached values, behaving differently.

schloerke, line 217: The whole function body could be something like...

```python
        conditions = {self._uid: uid}
        result = self.find_by(**conditions)
        if result is not None:
            return result
        raise ValueError(f"Object `\{ \"{self._uid}\": \"{ uid }\" \}` could not be found")
```
(untested)

src/posit/connect/resources.py
toph-allen: For my own understanding: What's the difference between these two?

toph-allen (approved):
looks good to me, pending resolving @schloerke’s comments :)

src/posit/connect/resources.py
schloerke: Technically nothing. They both resolve to `T | None`. 

I am on the stance of `Optional` should only be used as a parameter type as the parameter is "optional". But for typing, tell me it can have a `None` type.

src/posit/connect/resources.py
tdstein: This one is intentionally public to give the caller control over refreshing the collection from the server. Otherwise, there is not way for them to do so. I think `reload` flows better than `reset`. For example, `jobs.reload().find_by(hostname='example.com')` vs `jobs.reset().find_by(hostname='example.com')`

src/posit/connect/resources.py
tdstein: I think in either situation, we end up with conflicting ideas. 

If we always depend on `find_by`, we must fetch the entire collection before returning a single value, which would take significantly longer than a single GET request for the value.

Today, I think the obvious solution would be to **always** call the HTTP GET method to get the value from the server. This will sometimes be slightly slower than an in-memory list scan. But in reality, it's going to be a negligible difference. The weird edge case with this solution is when another process creates the value fetched by GET *after* the `_cache` is set. In this situation, the value returned by `find` will exist on the server but not in the _cache. This would probably warrant a cache invalidation. But that would take extra time to compute and may not be consistent behavior across all endpoints.

tl;dr - the speed up via `find_by` probably isn't worth the trouble. Classic over engineering.

src/posit/connect/resources.py
tdstein: Making it private is the right call. For some reason, single-word names for things spark joy in me. ✨  

I think the "conventional" name for this is `_is_cached` in most OOP books.

src/posit/connect/resources.py
tdstein: Unfortunately we have to use the explicit `Optional` and `Union` types in all cases for return types. The shorthand Union type operator was introduced in Python 3.10 and hasn't been backported to 3.9 or 3.8.

In 3.8, changing to `T | None` throws ` TypeError: unsupported operand type(s) for |: 'TypeVar' and 'NoneType'`.

This isn't the case for argument types. I don't know why yet, but it just works. My guess is that argument types aren't evaluated by the interpretation and Pyright is smart enough to deal with that fact. For example https://github.com/posit-dev/posit-sdk-py/blob/ac488c731d4a456555ef2ed6102cfaa3ca7b4d84/src/posit/connect/content.py#L320-L321 🤷🏻 

Side note; I've always thought of it as the inverse. Accept `T | None` and return `Optional[T]`. Or, in natural language, this method accepts `T` or `None` and sometimes returns `T`. Which stems from my background in Java, where the `T | None` syntax doesn't exist.

schloerke: By requiring the use of 

```
```

We can use `|` syntax on _not top level_ usage. 

```
Bad = int | str

class Ok:
    def method(self) -> int | str: ...
```

(This check is now made within ruff and will give a good suggested fix 😄 )

PR #316, opened 2024-10-28
build(uv): Lean on `uv` for all tasks within Makefile

schloerke:
## Changes

~~Breaking changes:~~
* ~~Drop python 3.8 support~~
	* ~~Main motivation: uv doesn't like to run quartodoc when python 3.8 is supported. However, python 3.8 is EOL, so it is fair to drop support.~~
	* Update: Workarounds were made to have quartodoc and uv play nicely. No remaining reason to drop 3.8

UV:
* Require usage of `.venv`
* Use `uv` for as many commands as possible
	* Every `make` command should be able to be run **without** requiring to activate the virtual environment
	* (Reduce adoption friction down to a minimum)
* Always install deps with `uv` before a python package command is run
* Remove requirements.txt and requirements-dev.txt. Instead, leverage pyproject.toml
	* Drop `make deps` and fold functionality into `make dev`
	* Currently waiting on `pip` to determine who groups are installed before `uv` implements group installation: https://github.com/astral-sh/uv/issues/8590#issuecomment-2439611796 . However, `uv` is fast that groups are not really needed for now.

GHA:
* Use `astral-sh/setup-uv` and `uv python install VERSION` to install `uv` and `python`
* Remove usage of `env: { UV_SYSTEM_PYTHON: true }`
* Drop cache from runs. `uv` is fast to install everything and nothing can get poisoned from prior runs. (Major pain to discover / debug)
* Test on python 3.13


Makefile
```diff
@@ -19,62 +19,63 @@ clean:
...
 	$(MAKE) -C ./docs
 
-ensure-uv:
+$(VIRTUAL_ENV):
+	$(UV) venv $(VIRTUAL_ENV)
+_ensure-uv-venv: _ensure-uv-cmd $(VIRTUAL_ENV)
+_ensure-uv-cmd:
 	@if ! command -v $(UV) >/dev/null; then \
-		$(PYTHON) -m ensurepip && $(PYTHON) -m pip install uv; \
+		$(PYTHON) -m ensurepip && $(PYTHON) -m pip install "uv >= 0.4.27"; \
```
tdstein, line 43: Since I have `uv` installed via Homebrew, I did not hit the `uv >= 0.4.27` version check. It took me a few minutes to realize that I needed to upgrade `uv` on my system.

I think this would be fixed by running `$(UV) pip install "uv >= 0.4.27"` after the if statement.

Makefile
```diff
@@ -19,62 +19,63 @@ clean:
...
-docs:
+docs: ensure-uv
 	$(MAKE) -C ./docs
 
-ensure-uv:
+$(VIRTUAL_ENV):
+	$(UV) venv $(VIRTUAL_ENV)
+_ensure-uv-venv: _ensure-uv-cmd $(VIRTUAL_ENV)
+_ensure-uv-cmd:
 	@if ! command -v $(UV) >/dev/null; then \
```
tdstein, line 42: Could this be a single command?

Makefile
```diff
@@ -19,62 +19,63 @@ clean:
...
-		$(PYTHON) -m ensurepip && $(PYTHON) -m pip install uv; \
+		$(PYTHON) -m ensurepip && $(PYTHON) -m pip install "uv >= 0.4.27"; \
 	fi
+ensure-uv: _ensure-uv-cmd _ensure-uv-venv
 
-fmt:
-	$(PYTHON) -m ruff check --fix
-	$(PYTHON) -m ruff format
+fmt: dev
```
tdstein, line 47: I have been hesitant to create these makefile target dependencies in the past due to the execution time overhead of running `dev` everytime. But it appears that the overhead is trival using this setup. Just to double check myself, can you think of any other reasons why we wouldn't want to run `dev` before each of these commands?

.github/workflows/ci.yaml
```diff
@@ -26,21 +21,16 @@ jobs:
...
-          - '3.8'
-          - '3.9'
-          - '3.10'
-          - '3.11'
-          - '3.12'
+          - "3.9"
+          - "3.10"
+          - "3.11"
+          - "3.12"
     steps:
```
tdstein, line 28: As mentioned elsewhere, we unfortunately cannot drop support for Python 3.8 at this time.

.github/workflows/ci.yaml
```diff
@@ -4,19 +4,14 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: 3.x
-          check-latest: true
-          cache: 'pip'
-      - run: make deps
+      - uses: astral-sh/setup-uv@v3
```
tdstein, line 12: It appears this action has caching support. https://github.com/astral-sh/setup-uv?tab=readme-ov-file#enable-caching

Did you have a chance to try it?

docs/Makefile
```diff
@@ -21,26 +21,36 @@ endif
...
 all: deps api build
 
-api:
-	$(QUARTODOC) build
-	$(QUARTODOC) interlinks
+ensure-dev:
+	$(MAKE) -C .. dev
+
+api: ensure-dev
+	@echo "::group::quartodoc interlinks"
```
tdstein, line 28: Smart. Just curious... does the `::group::task` nomenclature come from somewhere?

docs/Makefile
```diff
@@ -21,26 +21,36 @@ endif
...
+	$(UV) run $(QUARTODOC) build --verbose
+	@echo "::endgroup::"
 	cp -r _extensions/ reference/_extensions  # Required to render footer
 
-build:
-	CURRENT_YEAR=$(CURRENT_YEAR) \
-	PROJECT_VERSION=$(PROJECT_VERSION) \
-	$(QUARTO) render
+build: ensure-dev
+	source ../.venv/bin/activate && \
```
tdstein, line 37: Should this utilize `$(MAKE) -C .. ensure-uv` instead?

docs/Makefile
```diff
@@ -21,26 +21,36 @@ endif
...
 
 clean:
 	rm -rf _extensions _inv _site .quarto reference objects.json
 	find . -type d -empty -delete
 
-deps:
-	$(UV) pip install --upgrade pip -r requirements-site.txt
+_extensions/posit-dev/posit-docs/_extension.yml:
 	$(QUARTO) add --no-prompt posit-dev/product-doc-theme@v4.0.2
+_extensions/machow/interlinks/_extension.yml:
```
tdstein, line 48: Can we still enforce upgrades to Quartodoc this way?

Dockerfile
```diff
@@ -1,14 +1,10 @@
...
-ENV UV_SYSTEM_PYTHON=true
-
 RUN apt-get update && apt-get install -y make
 
 WORKDIR /sdk
 
-COPY Makefile pyproject.toml requirements.txt requirements-dev.txt vars.mk ./
-
-RUN --mount=type=cache,mode=0755,target=/root/.cache/pip make deps
+COPY Makefile pyproject.toml vars.mk ./
```
tdstein, line 7: Do we need to include `uv.lock` here?

tdstein (changes requested):
This is awesome! We'll need to adjust to continue supporting Python 3.8, but otherwise this looks great. It's wild how many new features `uv` has introduced since we originally migrated to it in August.

tdstein:
Also, you can remove the '!' from the conventional commit title. The '!' is only necessary for breaking changes affecting package consumers (e.g.,  a change requiring a major version bump). `build(uv): lean on `uv` for all Python tasks` would also be appropriate since this affects more than the Makefiles.

docs/Makefile
schloerke: The environment has to be active. Maybe we can get it to work with `$(UV) tool run --with ../ $(QUARTO) render`?

docs/Makefile
schloerke: No. Currently docs are not kept in version control. So GHA should pass as it is a clean environment.

I'm up for adding the quarto extensions folder though. This would resolve the issue and isn't much in files.

docs/Makefile
schloerke: It's a GHA thing. Gives it the little drop down to consolidate stdout/stderr within a single step. There can be many output groups per step.

docs/Makefile
schloerke: The `$(MAKE) -C .. ensure-uv` is performed within the `ensure-dev` target (so we'd only need to call it once for multiple targets)

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "posit-sdk"
 description = "Posit SDK for Python"
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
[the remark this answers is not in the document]
nealrichardson: Customers often have content running on older versions, and it's not (yet) trivial within Connect for them to "just" upgrade Python. So, if we stop supporting older versions in Connect, the effect is that it prevents customers from upgrading Connect because their content would stop working. We're working on a few things to make this better, but until we offer a better path for these customers, we're not dropping Python versions anymore.

The SDK doesn't necessarily need to follow what Connect does, but there are arguments for keeping it more permissive so that it is more widely usable by customers who haven't upgraded yet.

docs/Makefile
tdstein: Makes sense. Actually, since `make clean` clears out the extensions folder, it seems reasonable. My concern was for local development where these files are long lived and become out of date, but `make clean` is a decent escape hatch.

Makefile
schloerke: There is a hiccup in running it. But I believe the advantages outweigh the small slowdown for each run

Dockerfile
schloerke: I want to go with "no". It is possible that packages do not install the exact same way between local mac and docker linux.

If it is important to have the exact same setup, then we should commit uv.lock.

Makefile
schloerke: Yes.

Converting it to a `$(MAKE) $(VIRTUAL_ENV)` call after `uv` has been installed

pyproject.toml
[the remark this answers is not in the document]
schloerke: Thank you!

I'll add it back as it was more build tool driven, rather than package code requirements

Dockerfile
tdstein: It looks like it is operating system agnostic https://docs.astral.sh/uv/concepts/projects/#project-lockfile

In that case, we should be able to check it in and use it here.

docs/Makefile
tdstein: Whoa! That's awesome. TIL

pyproject.toml
```diff
@@ -22,7 +22,8 @@ dynamic = ["version"]
 dependencies = [
     "requests>=2.31.0,<3",
     "packaging",
-    "typing-extensions"
+    "typing-extensions",
+    "setuptools-scm;python_version>='3.12'",
```
tdstein, line 26: `setuptools-scm` should only be required for the build group. It is used to derive the version information for Git

Dockerfile
schloerke: I'm still on the "no" train. If it is important to resolve to the same packages, we should use stronger package version requirements.

(I'd agree for locked-down/reproducible environments, but for open source projects its nice to have some natural chaos / updates as users will not have access to our lock file)

Dockerfile
```diff
@@ -1,14 +1,10 @@
...
 
-ENV UV_SYSTEM_PYTHON=true
-
 RUN apt-get update && apt-get install -y make
 
 WORKDIR /sdk
 
-COPY Makefile pyproject.toml requirements.txt requirements-dev.txt vars.mk ./
-
-RUN --mount=type=cache,mode=0755,target=/root/.cache/pip make deps
```
tdstein, line 11: Since the `make dep` step has been combined with `make dev`, the iterative development process of integrations is quite a bit slower. On main, it took 7.3s to run `make -C integration build`, while on this branch, it took 13.1s. This isn't a huge difference, but it is noticeable since this is part of my development workflow. The reasoning may be a bit nuanced...

Line 11 creates a cached layer with the dependencies. When the source code is changed, it invalidates line 14. Subsequently, only line 16, which is the editable install, is recomputed. So, when doing development work on the source code, I can make changes without forcing Docker to reinstall all dependencies.


How difficult would it be to separate back out the editable install (`pip install -e .`) from the general dependency installation?

This isn't a blocker, but I would prefer the faster iterative build times.

Makefile
```diff
@@ -68,7 +68,7 @@ uninstall: ensure-uv
 
 version:
 	@$(MAKE) ensure-uv &>/dev/null
-	@$(UV) run --quiet python -m setuptools_scm
+	@$(UV) run --quiet --with "setuptools_scm" python -m setuptools_scm
```
tdstein, line 71: Nice!

Dockerfile
tdstein: Yeah, I had the same thought. I don't have strong opinions either way.

tdstein (approved):
🙌🏻 

I have one outstanding question regarding the integration tests, but regardless, this is awesome! I love the pragmatic use of `uv` in different places. It really shows the power of the tooling.

Dockerfile
schloerke: 🥳 Solved with `$(UV) sync --frozen --no-install-project`

Installs given the lock file (that is copied over). Only install the project dependencies, but not the `posit` package (which needs setuptools-scm and `./src` to be compiled).

PR #320, opened 2024-11-04
refactor: removes cache layer from active sequence

tdstein:
Removes the cache layer from finder methods to reduce the complexity of propagating query parameters.

When passing query parameters through finder methods, it is difficult to manage the cache if we wish to have it reflect the user's query state. For example, when calling `packages.find_by(name='posit-sdk)',` what should `packages._data` return on the next call?


tdstein:
Made `fetch` public as a more verbose way of getting the list. This could also be named `aslist` or something similar.

schloerke (approved):
I'd like to explore more of possibly removing the caching layer completely in a followup. 

But I like where this is going!

PR #323, opened 2024-11-08
fix: remove all warnings from attribute access

tdstein:


tests/posit/connect/oauth/test_integrations.py
```diff
@@ -1,61 +1,11 @@
...
-from posit.connect.oauth.integrations import Integration
 
 from ..api import load_mock  # type: ignore
 
 
-class TestIntegrationAttributes:
-    @classmethod
-    def setup_class(cls):
-        guid = "22644575-a27b-4118-ad06-e24459b05126"
-        fake_item = load_mock(f"v1/oauth/integrations/{guid}.json")
```
schloerke, line 17: With both tests being removed, should we remove the mock files?

tdstein: The mocks are also used in other tests below.

schloerke (approved):
Thank you!

PR #329, opened 2024-11-12
fix: raise AttributeError instead of returning None on __getattr__

tdstein:
The` __getattr__` method should always raise an `AttributeError` if the value does not exist instead of returning `None`. The existing implementation causes issues when converting to a Pandas `DataFrame`.

This error was introduced in the 0.5.0 release.

### DIFF
--- a/src/posit/connect/resources.py
+++ b/src/posit/connect/resources.py
@@ -40,7 +40,7 @@ class Resource(dict):
                 stacklevel=2,
             )
             return self[name]
-        return None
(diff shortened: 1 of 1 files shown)


zackverham:
Seems simple enough! Any reason to try to capture this behavior in a test somewhere? Its generic enough that the answer might be no.

zackverham (approved):

tdstein:
> Seems simple enough! Any reason to try to capture this behavior in a test somewhere? Its generic enough that the answer might be no.

Eh, that code block will be gone soon enough. Probably not worth the effort at this point.

PR #330, opened 2024-11-12
[hold] refactor: Merge Active and Api code together

### LINKED ISSUE #324: Merge _api.py and resource.py in code
TODO - make an issue to port this functionality to the Resource class.

_Originally posted by @tdstein in https://github.com/posit-dev/posit-sdk-py/pull/300#discussion_r1833107446_

schloerke:
Fixes #324

-------------------------------------------------------

- [x] Add to readme file more in-depth information: https://github.com/posit-dev/posit-sdk-py/pull/330/files#diff-97a404af0cd76efe034f1917316121f8ff4be4a51655e5bc16b65f5800cdb2b7
- [ ] Make PR in rstudio/connect to address the no "alter in place" updates

Goals:
* Turn entity classes into read only dictionaries. Prevents users from updating data in place. All `.update()`-like methods now return a new object.
* Minimize public parameters such as `.params: ResourceParameters` (Rename to `._ctx: Context`
	* Inherit from `Context` to extend it with more information rather than expanding parameters. By doing so, the names and usage are consistent. (Hard to distinguish which GUID `guid` represents )
	* Leverage the `ContextP[ContextT]` class where possible to give `._ctx: ContextT` a type. 
* Minimize the locations of API path. Resolve by placing the calling class. 
	* Classes that need to fully fetch their information may provide the bare minimum information for the `ActiveDict` class to request the data for initialization. This reduces multiple duplications of logic to a single spot
* Attempt to use `ApiCallMixin` within classes to reduce how urls are constructed and processed. Given a `.ctx: Context` and `._path: str`, many calling methods can be handled under the hood.

Note - Classes that do not follow the read-only approach: 
* `EnvVars`
* `ContentItemVanityMixin`


src/posit/connect/jobs.py
```diff
@@ -150,7 +150,8 @@ def _create_instance(self, path: str, /, **attributes: Any) -> Job:
 
     class _FindByRequest(TypedDict, total=False):
         # Identifiers
-        id: Required[str]
+        # TODO-barret-Q: Is this change correct? It use to be required, but the docs say not-required
+        id: NotRequired[str]
```
schloerke, line 154: Is this change correct? It use to be required, but the docs say not-required

src/posit/connect/oauth/associations.py
```diff
@@ -1,78 +1,106 @@
...
+        return f"v1/content/{content_guid}/oauth/integrations/associations"
 
-    def find(self) -> List[Association]:
+    def __init__(self, ctx: ContentItemContext) -> None:
+        super().__init__()
+        self._ctx = ctx
+        self._path = self._api_path(ctx.content_guid)
+
+    # TODO-barret-future-q: Should this be inherited from ActiveFinderSequence? (It would add find_by)
+    def find(self) -> list[Association]:
```
schloerke, line 79: Should this be inherited from ActiveFinderSequence? (It would add find_by) . We could have `.find()` and `.find_by()` be separated out into two `ActiveFinder*` classes.

src/posit/connect/oauth/integrations.py
```diff
@@ -1,56 +1,82 @@
...
-        """Update the OAuth integration."""
 
-    def update(self, *args, **kwargs) -> None:
-        """Update the OAuth integration."""
-        body = dict(*args, **kwargs)
-        url = self.params.url + f"v1/oauth/integrations/{self['guid']}"
-        response = self.params.session.patch(url, json=body)
-        super().update(**response.json())
+# TODO-barret-future-q; Should this auto retrieve? If so, it should inherit from ActiveSequence
+class Integrations(ApiCallMixin, ContextP[Context]):
```
schloerke, line 69: I'm really enjoying the clarity of when data is retrieved from `.find()` / `.get()` / `.all()` methods. And it does take the ambiguity out of _when_ data is being retrieved. 

I'd be up for dropping the "download all list-like data on initialization" in favor of the retrieval methods (e.g. `.find()` etc.) . This would also mean that the classes wouldn't be `list-like` would require the user to retrieve before calling any list like methods (ex: `len(jobs.all())` and not `len(jobs)` which can change over time.

src/posit/connect/oauth/oauth.py
```diff
@@ -4,26 +4,30 @@
...
-class OAuth(Resources):
-    def __init__(self, params: ResourceParameters, api_key: str) -> None:
-        super().__init__(params)
+class OAuth(ContextP[Context]):
+    def __init__(self, ctx: Context, api_key: str) -> None:
+        super().__init__()
+        self._ctx = ctx
+
+        # TODO-barret-q: Is this used?
         self.api_key = api_key
```
schloerke, line 19: I couldn't see where this is used. Just wanted to be sure I didn't lose this one as it is related to auth.

src/posit/connect/oauth/sessions.py
```diff
@@ -2,34 +2,101 @@
...
+        guid = assert_guid(kwargs.get("guid"))
+        path = self._api_path(guid)
+
+        # Only fetch data if `kwargs` only contains `"guid"`
+        get_data = len(kwargs) == 1
+
+        super().__init__(ctx, path, get_data, **kwargs)
+
+    # TODO-barret-future-q: Should this be destroy?
     def delete(self) -> None:
```
schloerke, line 69: Should this be `.delete()` or `.destroy()`?

src/posit/connect/oauth/sessions.py
```diff
@@ -2,34 +2,101 @@
...
         self,
         *,
         all: Optional[bool] = ...,
     ) -> List[Session]: ...
 
+    # TODO-barret-future-q: Should this be `.find_by()`?
     @overload
     def find(self, **kwargs) -> List[Session]: ...
 
     def find(self, **kwargs) -> List[Session]:
```
schloerke, line 95: No. `all: bool` is a query parameter. Instead, it should be `**kwargs: Unpack[Sessions._FindAttrs]`

schloerke:
After conversations with Taylor, we are not going to merge this PR. However, there are good ideas within it. 

Will keep the PR open until issues are made and/or components pulled out.

tdstein:
Closing this change. It will continue to exist in the pull request history for future reference.

PR #332, opened 2024-11-12
refactor: Remove `dict` from parent class of Context

schloerke:


src/posit/connect/context.py
```diff
@@ -23,25 +27,27 @@ def wrapper(instance: ContextManager, *args, **kwargs):
...
+        endpoint = self.url + "server_settings"
+        response = self.session.get(endpoint)
+        result = response.json()
+        value = self._version = result.get("version")
         return value
 
     @version.setter
     def version(self, value):
-        self["version"] = value
+        self._version = value
```
tdstein, line 50: How about...

```python
class Context:
    def __init__(self, session: requests.Session, url: Url):
        self.session = session
        self.url = url

    @property
    def version(self) -> str | None:
        if not hasattr(self, "_version"):
            endpoint = self.url + "server_settings"
            response = self.session.get(endpoint)
            result = response.json()
            self._version: str | None = result.get("version")

        return self._version

    @version.setter
    def version(self, value: str | None):
        self._version = value
```

`hasattr` is more common than checking the symbol table directly. And specifying the types on the setter and on L45 avoids the empty declaration in `__init__`

tdstein (approved):
👍🏻

Should this be merged into `main`? Seems like it would be beneficial regardless the outcome of https://github.com/posit-dev/posit-sdk-py/tree/schloerke/324-merge-api-and-resource

Issue #338, opened 2024-11-15
Add Revoking Access from Group


schloerke:
Related Recipe Issue: https://github.com/rstudio/connect/issues/28826

Current recipe code: https://docs.posit.co/connect/cookbook/content-access-controls/revoking-access-from-a-group/

```python
from posit import connect

#### User-defined inputs ####
# 1. specify the guid for the content item
content_guid = "CONTENT_GUID_HERE"
# 2. specify either the guid or name for the group to be removed (group_name will be used if group_guid is blank)
group_guid = ""
group_name = "GROUP_NAME_HERE"
############################

client = connect.Client()

# search by group_name to find the group_guid if blank
if not group_guid and group_name:
    group_match = client.get("/v1/groups", params={"prefix": group_name}).json()
    if not group_match["results"]:
        raise Exception("Invalid group name")
    elif len(group_match["results"]) != 1:
        raise Exception("More than one group name found, ensure you enter a unique name")
    else:
        group_guid = group_match["results"][0]["guid"]
elif not group_name:
    raise Exception("Either group_guid or group_name must be specified")

# For the specified content item remove the desired user
for perm in client.content.get(content_guid).permissions.find():
    if perm.principal_guid == group_guid:
        perm.delete()

# Confirm new permissions
client.content.get(content_guid).permissions.find()
```

schloerke:
Proposed recipe

```python
from posit import connect

#### User-defined inputs ####
# 1. specify the guid for the content item
content_guid = "CONTENT_GUID_HERE"
# 2. specify either the guid or name for the group to be removed (group_name will be used if group_guid is blank)
principle_guid = "USER_OR_GROUP_GUID_HERE"
group_name_prefix = "GROUP_NAME_PREFIX_HERE"
############################

client = connect.Client()

# Remove a single permission by group_guid
client.content.get(content_guid).permissions.delete(principle_guid)

# Remove all groups with a matching prefix name
groups = client.groups.find(prefix=group_name_prefix)
client.content.get(content_guid).permissions.delete(*groups)
```

tdstein:
Nice! Let's use `destroy` instead of `delete` as the method name. But this looks great!

tdstein:
[...]

PR #339, opened 2024-11-15
feat: Add `Permissions.delete(*permissions)` method

schloerke:
Fixes: https://github.com/posit-dev/posit-sdk-py/issues/338
Related: https://github.com/rstudio/connect/issues/28826

#### Other changes
* ~~`Permission.delete()` deprecated for `Permission.destroy()`~~


integration/tests/posit/connect/test_users.py
```diff
@@ -29,8 +33,8 @@ def test_lock(self):
         assert len(self.client.users.find(account_status="locked")) == 0
 
     def test_count(self):
-        # aron, bill, cole, and me
-        assert self.client.users.count() == 4
+        # aron, bill, cole, and me (and existing user)
+        assert self.client.users.count() == 3 + self.existing_users
```
jonkeane, line 37: Could you say more about why `3 + self.existing_users` is preferred to the hard code?

It feels a little off that this blends something from setup with a test here: if `count()` were to not work at all, that would make the setup always fail. And if we wanted to test that `client.users.count(); client.users.create(); client.users.count()` increments like we expect, that should be in a test block, right?

src/posit/connect/_deprecated.py
```diff
@@ -0,0 +1,23 @@
+import warnings
+
+
+# From https://github.com/posit-dev/py-shiny/blob/f6b92d8cf49a90f3b3dbb636cd6d7fdeee244cfd/shiny/_deprecated.py#L15C1-L28C1
+# Create our own warning class instead of using built-in DeprecationWarning, because we
+# want to be able to control display of these messages without interfering with the
+# user's control of DeprecationWarning.
+# 2024-11: Change base class to DeprecationWarning
```
jonkeane, line 8: Is this a behavior we want to override? 

There does appear to be some churn in how silent these are: https://docs.python.org/3/library/warnings.html#default-warning-filter but it smells a bit funny to me to wrap around core language functionality / expectations. Not saying we _absolutely shouldn't_ but this might be a larger conversation than this PR?

tests/posit/connect/api.py
```diff
@@ -2,10 +2,8 @@
```
jonkeane, line 2: Are the changes in this file needed for what you're doing? They don't seem connected, but maybe I'm missing something

src/posit/connect/permissions.py
```diff
@@ -137,3 +148,86 @@ def get(self, uid: str) -> Permission:
         url = self.params.url + path
         response = self.params.session.get(url)
         return Permission(self.params, **response.json())
+
+    def destroy(self, *permissions: str | Group | User | Permission) -> list[Permission]:
+        """Destroy supplied content item permissions.
```
jonkeane, line 153:
```suggestion
        """Remove supplied content item permissions.
```

This might speak to our users more than `Destroy`?

src/posit/connect/permissions.py
```diff
@@ -137,3 +148,86 @@ def get(self, uid: str) -> Permission:
...
+
+        Removes all provided permissions from the content item's permissions.
+
+        Parameters
+        ----------
+        *permissions : str | Group | User | Permission
+            The content item permissions to remove. If a `str` is received, it is compared against
+            the Permissions' `principal_gruid`. If a `Group`, `User`, or `Permission` is received,
```
jonkeane, line 161:
```suggestion
            the Permissions' `principal_guid`. If a `Group`, `User`, or `Permission` is received,
```

typo?

src/posit/connect/permissions.py
```diff
@@ -137,3 +148,86 @@ def get(self, uid: str) -> Permission:
...
+            the `guid` is used and compared against the Permissions' `principal_guid`. If a
+            `Permission` is received, the `principal_guid` is used. Note, only the associated permissions will be destroyed; Any users or groups provided will remain.
```
jonkeane, line 163: I don't think I follow what this is trying to say? Looking at this + RTFS lower: I think that `Permission` might be wrong in the line above. And then it's saying that if you have a `Group` or a `User`, then the `guid` field from that object will be the `guid` that is used to remove the permission. If you have a `Permission` then the `permission_guid` field is what is used to remove. Is that right?

I wonder if that's a level of detail that is helpful for users or if it's just distracting?

src/posit/connect/permissions.py
```diff
@@ -137,3 +148,86 @@ def get(self, uid: str) -> Permission:
...
+                raise TypeError(
+                    f"destroy() expected argument type 'str', 'Group', 'Permission' or 'User', but got '{type(arg).__name__}'",
+                )
+            principal_guids.add(principal_guid)
+
+        destroyed_permissions: list[Permission] = []
+        for permission in self.find():
+            if permission["principal_guid"] in principal_guids:
+                permission.destroy()
+                destroyed_permissions.append(permission)
```
jonkeane, line 231: Am I following correctly that this will silently skip any Users/Groups/Permissions that are in the args, but aren't already on the piece of content? That's probably the right behavior, but should we add that to the documentation? Are there edge cases we want to catch and warn about (i.e. there are no matches at all)?

src/posit/connect/permissions.py
```diff
@@ -137,3 +148,86 @@ def get(self, uid: str) -> Permission:
...
+                principal_guid = arg
+            elif isinstance(arg, Group):
+                principal_guid: str = arg["guid"]
+            elif isinstance(arg, User):
+                principal_guid: str = arg["guid"]
+            elif isinstance(arg, Permission):
+                principal_guid: str = arg["principal_guid"]
+            else:
+                raise TypeError(
+                    f"destroy() expected argument type 'str', 'Group', 'Permission' or 'User', but got '{type(arg).__name__}'",
```
jonkeane, line 223:
```suggestion
                    f"destroy() expected argument type 'str', 'User', 'Group', or 'Permission' but got '{type(arg).__name__}'",
```

I know it's not alphabetic order, but I find what's there to be a strange order. This order is in order of most-likely to be used / most likely to be a concept someone is familiar with.

jonkeane (changes requested):
Thanks for this, a few comments for you. 

Would you mind adding more detail (can be only in the comments of the PR) about why we need to deprecate `delete` and transition to `destroy` for this? I am not necessarily opposed, but from the code itself it doesn't seem to be required and it would be good to have a papertrail of why we did that.

src/posit/connect/permissions.py
```diff
@@ -2,16 +2,27 @@
...
+if TYPE_CHECKING:
+    from .groups import Group
+    from .users import User
+
 
 class Permission(Resource):
     def delete(self) -> None:
-        """Delete the permission."""
+        """[Deprecated] Delete the permission."""
+        warn_deprecated_and_removed_in_future("Please use .destroy() instead of .delete().")
```
nealrichardson, line 20: We aren't at 1.0 yet, so do we need to be so cautious about deprecation cycles?

schloerke: My thought was that the deprecations could be gutted when v1 comes out.

integration/tests/posit/connect/test_content_item_permissions.py
```diff
@@ -0,0 +1,86 @@
...
+        self.content.permissions.create(
+            principal_guid=self.user_aron["guid"],
+            principal_type="user",
+            role="viewer",
+        )
+        self.content.permissions.create(
+            principal_guid=self.group_friends["guid"],
+            principal_type="group",
+            role="owner",
+        )
```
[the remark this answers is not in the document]
schloerke: Added in https://github.com/posit-dev/posit-sdk-py/pull/343

integration/tests/posit/connect/test_users.py
schloerke: In the test file above (integration/tests/posit/connect/test_content_item_permissions.py), I want to use a couple users to test the permissions. So I added them there to have it be self contained. (We could move the users to an earlier location, but enforcing that the users exist for any subset of testing seemed like a big hassle.)

In slack, Aron stated that users can not be deleted from within Connect. So if another user is added in a different test, then the total number of users in this test will be larger than 4 (aron, bill, cole, and me). 

> It feels a little off that this blends something from setup with a test here

Agreed. But each of the other tests require the three user objects to already exist. 

We could move the `.create()` test to another test class and retrieve the users in this class to avoid crossing setup / testing. Using a subset of tests won't work well if trying to just test this situation but that's ok.

tests/posit/connect/api.py
schloerke: I had added them in an earlier PR and Taylor had pushed back. I've come to agree that the type requirements should be relaxed.

So yes, unrelated to this PR, but useful.

src/posit/connect/permissions.py
schloerke: Correct, mismatches will be silent. 

I'll add that to the docs.

I'd like to not submit warnings if possible. My interpretation is that at the end of the function, all associated provided permissions will not exist. The permissions that were removed are returned so users can make their own assertions / warnings.

src/posit/connect/permissions.py
schloerke: Simplified above

src/posit/connect/_deprecated.py
schloerke: Issue created #344 for followup work

integration/tests/posit/connect/test_users.py
jonkeane: > In slack, Aron stated that users can not be deleted from within Connect. So if another user is added in a different test, then the total number of users in this test will be larger than 4 (aron, bill, cole, and me).


_nods_ right right, that's true, but also true if we have `4` there or if we do `3 + self.existing_users`, right? We would need to do `4 + self.existing_users` or `5 + self.existing_users` if we add more users. Or am I missing something?

src/posit/connect/permissions.py
nealrichardson: I don't feel strongly, just that we probably have only a handful of users, if any, of the existing code, so it's probably not worth much trouble. Also, for recipes running on Connect, I don't think we meaningfully expose DeprecationWarnings to the owners (they probably show up in the logs, but you'd have to know to look for them), so the practical value is limited unless/until that is surfaced better.

schloerke: Great! I'll remove the methods and deprecation codes. 😁😁

integration/tests/posit/connect/test_users.py
schloerke: In this test we only add 3x users: `aron`, `bill`, and `cole`. So only if we add more users within _this_ test will we need to change `3` to a new value.

src/posit/connect/permissions.py
```diff
@@ -137,3 +141,87 @@ def get(self, uid: str) -> Permission:
...
+        from .users import User
+
+        if len(permissions) == 0:
+            raise ValueError("Expected at least one `permission` to remove")
+
+        principal_guids: set[str] = set()
+
+        for arg in permissions:
+            if isinstance(arg, str):
+                principal_guid = arg
```
jonkeane, line 208: Do we have other facilities for checking if a string is the right shape to be a GUID? That might be nice to check here since someone might try passing `client.content.get(content_guid).permissions.destroy("my group name")` or the like?

tests/posit/connect/test_permissions.py
```diff
@@ -262,3 +265,81 @@ def test(self):
...
+            fake_permission["principal_guid"],
+            # Extract info from User, Group, Permission
+            user_to_remove,
+            group_to_remove,
+            permission_to_remove,
+        )
+
+        # Assert bad input value
+
+        with pytest.raises(TypeError):
```
jonkeane, line 325:
```suggestion
        # Assert bad input value
        with pytest.raises(TypeError):
```

Is this extra whitesapce? Or is it intentional?

jonkeane (approved):
Thanks for cleaning this up, I like where this is. A few minor comments, mostly style / docs.

src/posit/connect/permissions.py
```diff
@@ -137,3 +141,87 @@ def get(self, uid: str) -> Permission:
...
+            elif isinstance(arg, Group):
+                principal_guid: str = arg["guid"]
+            elif isinstance(arg, User):
+                principal_guid: str = arg["guid"]
```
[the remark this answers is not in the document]
schloerke: Yes!

src/posit/connect/permissions.py
schloerke: No. We do not have the functionality. 

I believe the thought process is to have the server return an error and we display the error.

src/posit/connect/permissions.py
```diff
@@ -137,3 +141,87 @@ def get(self, uid: str) -> Permission:
...
+        """Remove supplied content item permissions.
+
+        Removes all provided permissions from the content item's permissions. If a permission isn't
+        found, it is silently ignored.
+
+        Parameters
+        ----------
+        *permissions : str | Group | User | Permission
+            The content item permissions to remove. If a `str` is received, it is compared against
+            the `Permissions`' `principal_guid`. If a `Group` or `User` is received, the associated
```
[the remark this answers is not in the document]
schloerke:
```suggestion
            The content item permissions to remove. If a `str` is received, it is compared against
            the `Permissions`'s `principal_guid`. If a `Group` or `User` is received, the associated
```

src/posit/connect/permissions.py
schloerke: Should we instead change the approach of this method to **always** submit the delete API calls to the server? (rather than only the permissions we know exist.)

Then the server will always display an error. And we wouldn't need to do any validation on our end.

jonkeane: AAAH right on line 223 it will silently be ignore if it's anything other than a matching GUID. HMMM I don't have a strong intuition which direction would be better here, pre-checking or always sending and failing. There certainly are complications with either.

jonkeane: > Should we instead change the approach of this method to always submit the delete API calls to the server? (rather than only the permissions we know exist.)

For posterity, did you end up implementing this or does it still (silently) ignore non-GUIDs (or really anything that doesn't match a GUID already present if it's a string)?

schloerke: It still silently ignores non existent permissions

Issue #342, opened 2024-11-19
Add users and groups to content item permissions


schloerke:
Related: https://github.com/rstudio/connect/issues/28825 ; https://docs.posit.co/connect/cookbook/content-access-controls/granting-access-to-a-group/
Related: Granting Access to a User: https://docs.posit.co/connect/cookbook/content-access-controls/granting-access-to-a-user/

#### Current Group recipe:

```python
from posit import connect

# 1. specify the guid for the content item
content_guid = "CONTENT_GUID_HERE"
# 2. specify either the guid or name for the group to be added (group_name will be used if group_guid is blank)
group_guid = ""
group_name = "GROUP_NAME_HERE"
# 3. specify if the group should be added as a "viewer" or "owner" (collaborator)
access_type = "viewer"

client = connect.Client()

# search by group_name to find the group_guid if blank
if not group_guid and group_name:
    group_match = client.get("/v1/groups", params={"prefix": group_name}).json()
    if not group_match["results"]:
        raise Exception("Invalid group name")
    elif len(group_match["results"]) != 1:
        raise Exception("More than one group name found, ensure you enter a unique name")
    else:
        group_guid = group_match["results"][0]["guid"]
elif not group_name:
    raise Exception("Either group_guid or group_name must be specified")

# For the specified content item add the desired group
client.content.get(content_guid).permissions.create(
    principal_guid=group_guid,
    principal_type="group",
    role=access_type,
    )

# Confirm new permissions
client.content.get(content_guid).permissions.find()
```

#### Current User recipe:

```python
from posit import connect

# 1. specify the guid for the content item
content_guid = "CONTENT_GUID_HERE"
# 2. specify either the guid or username for the user being added (username will be used if user_guid is blank)
user_guid = ""
username = "USERNAME_HERE"
# 3. specify if the user should be added as a "viewer" or "owner" (collaborator)
access_type = "viewer"

client = connect.Client()

# search by username to find the user_guid if blank
if not user_guid and username:
    user_match = client.users.find(prefix=username)
    if not user_match:
        raise Exception("Invalid username")
    elif len(user_match) != 1:
        raise Exception("More than one username found, ensure you enter a unique name")
    else:
        user_guid = user_match[0]["guid"]
elif not username:
    raise Exception("Either user_guid or username must be specified")

# For the specified content item add the desired user
client.content.get(content_guid).permissions.create(
    principal_guid=user_guid,
    principal_type="user",
    role=access_type,
)

[...]

PR #343, opened 2024-11-19
feat: Add `Permission.create(*args:User | Group)` support

schloerke:
Fixes https://github.com/posit-dev/posit-sdk-py/issues/342
Related: https://github.com/rstudio/connect/issues/28825

- [x] Unit testing
	- [x] Wait until #339 is merged to build out 
- [x] Integration testing
	- Not needed as cosmetic unit testing should cover it

### DIFF
--- a/src/posit/connect/permissions.py
+++ b/src/posit/connect/permissions.py
@@ -91,8 +91,8 @@ class Permissions(Resources):
         Returns
         -------
         Permission | List[Permission]
-            Returns a `Permission` when the kwargs: `principal_guid`, `principal_type`, and `role`
-            are used. Returns a `list[Permission]` when `*args` are used.
+            Returns a `Permission` when the kwargs: `principal_guid` and `principal_type` are used.
+            Returns a `list[Permission]` when `*args` are used.
 
         Examples
         --------
--- a/src/posit/connect/permissions.py
+++ b/src/posit/connect/permissions.py
@@ -67,11 +67,20 @@ class Permissions(Resources):
         return len(self.find())
 
     @overload
-    def create(self, *, principal_guid: str, principal_type: str, role: str) -> Permission:
+    def create(self, *, principal_guid: str, principal_type: str, role: str) -> Permission: ...
+
+    @overload
+    def create(self, *args: User | Group, role: str) -> list[Permission]: ...
+
+    def create(self, *args: User | Group, **kwargs) -> Permission | list[Permission]:
         """Create a permission.
 
         Parameters
         ----------
+        *args : User | Group
+            The principal users or groups to add.
+        role : str
+            The principal role. Currently only `"viewer"` and `"owner"` are supported.
         principal_guid : str
             User guid or Group guid.
         principal_type : str
@@ -81,23 +90,9 @@ class Permissions(Resources):
 
         Returns
         -------
-        Permission
-        """
-
-    @overload
-    def create(self, *args: User | Group, role: str) -> list[Permission]:
-        """Create a permission.
-
-        Parameters
-        ----------
-        *args : User | Group
-            The principal users or groups to add.
-        role : str
-            The principal role. Currently only `"viewer"` and `"owner"` are supported.
-
-        Returns
-        -------
-        Permission
+        Permission | List[Permission]
+            Returns a `Permission` when the kwargs: `principal_guid`, `principal_type`, and `role`
+            are used. Returns a `list[Permission]` when `*args` are used.
 
         Examples
         --------
@@ -141,8 +136,6 @@ class Permissions(Resources):
         content_item.permissions.find()
         ```
         """
-
-    def create(self, *args: User | Group, **kwargs) -> Permission | list[Permission]:
         if len(args) > 0:
             # Avoid circular imports
             from .groups import Group
@@ -232,7 +225,7 @@ class Permissions(Resources):
         Returns
         -------
         list[Permission]
-            The removed permissions. If a permission is not found, there is nothing to remove and 
+            The removed permissions. If a permission is not found, there is nothing to remove and
             it is not included in the returned list.
 
         Examples
--- a/src/posit/connect/permissions.py
+++ b/src/posit/connect/permissions.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, List, overload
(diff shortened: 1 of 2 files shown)


schloerke:
@jonkeane Looking for approval on the function implementation. I'll add unit testing after #339 is merged and ask again for a review before merging.

src/posit/connect/permissions.py
```diff
@@ -69,30 +73,104 @@ def create(self, *, principal_guid: str, principal_type: str, role: str) -> Perm
...
+            The principal role. Currently only `"viewer"` and `"owner"` are supported
 
         Returns
         -------
         Permission
         """
 
     @overload
-    def create(self, **kwargs) -> Permission:
+    def create(self, *args: User | Group, role: str) -> list[Permission]:
```
toph-allen, line 88: Merge the two overloaded documentation strings into one

src/posit/connect/permissions.py
```diff
@@ -69,30 +73,104 @@ def create(self, *, principal_guid: str, principal_type: str, role: str) -> Perm
...
+        if len(args) > 0:
+            # Avoid circular imports
+            from .groups import Group
+            from .users import User
+
+            for arg in args:
+                if not isinstance(arg, (User, Group)):
+                    raise TypeError(f"Invalid argument type: {type(arg)}")
+            if "principal_guid" in kwargs:
+                raise ValueError("'principal_guid' can not be defined with `*args` present.")
```
toph-allen, line 155: Check for `principal_type` as well.

toph-allen (approved):
Pair reviewed. Approved given the small doc changes :)

src/posit/connect/permissions.py
```diff
@@ -67,36 +67,105 @@ def count(self) -> int:
...
         -------
-        Permission
-        """
-
-    @overload
-    def create(self, **kwargs) -> Permission:
-        """Create a permission.
+        Permission | List[Permission]
+            Returns a `Permission` when the kwargs: `principal_guid`, `principal_type`, and `role`
+            are used. Returns a `list[Permission]` when `*args` are used.
```
jonkeane, line 95: Returning either an object of a specific type or a list of those objects feels a little weird as an interface here. Is that a common approach for things like these? It certainly would make programming with the SDK more finicky because one would have to remember when it'll be a list and when it won't, yeah?

schloerke: The `@overload`s should handle the typing interface. 

If this is a road we don't want for the package, I'll need to back it out of the other PRs as well.

schloerke, line 95:
```suggestion
            Returns a `Permission` when the kwargs: `principal_guid` and `principal_type` are used.
            Returns a `list[Permission]` when `*args` are used.
```

src/posit/connect/permissions.py
```diff
@@ -67,36 +67,105 @@ def count(self) -> int:
         return len(self.find())
 
     @overload
-    def create(self, *, principal_guid: str, principal_type: str, role: str) -> Permission:
+    def create(self, *, principal_guid: str, principal_type: str, role: str) -> Permission: ...
+
+    @overload
+    def create(self, *args: User | Group, role: str) -> list[Permission]: ...
+
+    def create(self, *args: User | Group, **kwargs) -> Permission | list[Permission]:
```
amol-, line 75: While I understand that changing return type of `create()` to always be `list[Permission]` is not wise from backward compatibility point of view, but I think that having a function switching behaviour and return type based on the arguments can be confusing. Especially in the context of type hinting.

Why not make a `create()` and `create_many()` functions?
`create()` could be a simple wrapper on top of `create_many()` and so it wouldn't add any additional maintenance cost.
At that point each one of the two can have a clear and unique signature and even from autocompletion point of view for a user it will be more obvious to see that they have two choices.

toph-allen: Pairing with @schloerke through a review of this, and we're discussing this.

This seems like a really reasonable approach. Question:

Should `create()` adopt user and group support, so you can call `.create(my_user, role="viewer")`?

schloerke: If `.create()` adopted user/group support, I'd be fine not implementing `.create_many()`

```python
    def create(self, principal: User | Group, /, **kwargs: Unpack[CreateArgs]) -> Permission:
```

toph-allen: Then you could just use it as in a list comprehension!

src/posit/connect/permissions.py
```diff
@@ -67,36 +67,105 @@ def count(self) -> int:
...
+        # New permission role
+        role = "viewer"  # Or "owner"
+
+        # Example groups and users
+        groups = client.groups.find(prefix="GROUP_NAME_PREFIX_HERE")
+        user = client.users.get("USER_GUID_HERE")
+        users = [user]
+
+        # Add many group and user permissions with the same role
+        content_item.permissions.create(*groups, *users, role=role)
```
toph-allen, line 114: If you make the above change, you could replace this with an example using a list comprehension on a list of groups and/or users.

src/posit/connect/permissions.py
schloerke: Oooooo. I like our suggestion below that we keep the same return type of `Permission` but essentially `*args` turns into `principal: User | Group` as a position only argument.

src/posit/connect/permissions.py
schloerke: Settled on :

```python
    @overload
    def create(self, /, *, principal_guid: str, principal_type: str, role: str) -> Permission: ...

    @overload
    def create(self, principal: User | Group, /, *, role: str) -> Permission: ...
```

😃😃

src/posit/connect/permissions.py
schloerke: I've backed out implementations that "handle arrays of information because why not" in `User.groups.add(group)`, `Group.members.add(user)`, and `Permissions.destroy(permission)` in [`ab5a4d5` (#341)](https://github.com/posit-dev/posit-sdk-py/pull/341/commits/ab5a4d5364865e8740889ed8a6287859101993ec)

This way if its a single API call under the hood, things are more explicit and less magic. (No need to wrap the for loop and then insert confusion / misdirection.) If a day comes where the Connect API handles array inputs / results, then I'm up for supporting it within the SDK.

schloerke:
Reviewed in person w/ @toph-allen . Merging

PR #348, opened 2024-12-02
feat: Add client credential oauth integration support + related databricks helpers to SDK

zackverham:
This PR adds a new function to the oauth tooling. It knows how to craft a request against the credential exchange endpoint that can interface with client credential flows. This supports service-account-like oauth interactions. 

Connect sets an environment variable called `CONNECT_CONTENT_SESSION_TOKEN` which can be used by the executing content when making the request. This means that the syntax for using this new credential exchange function is slightly different. 

This PR also adds the databricks extensions needed to use this new oauth exchange flow when interfacing with databricks resources. 

This feature won't land in Connect until the upcoming release gets published - I don't know how we typically queue up work like this in sync with product releases.

### DIFF
--- a/src/posit/connect/external/databricks.py
+++ b/src/posit/connect/external/databricks.py
@@ -8,11 +8,11 @@ https://github.com/databricks/databricks-sdk-py#interface-stability
 import abc
 from typing import Callable, Dict, Optional
 
+import requests
+
 from ..client import Client
 from ..oauth import Credentials
 from .external import is_local
-
-import requests
 
 POSIT_OAUTH_INTEGRATION_AUTH_TYPE = "posit-oauth-integration"
 POSIT_LOCAL_CLIENT_CREDENTIALS_AUTH_TYPE = "posit-local-client-credentials"
@@ -81,14 +81,15 @@ def _get_auth_type(local_auth_type: str) -> str:
 
     return POSIT_OAUTH_INTEGRATION_AUTH_TYPE
 
+
 class PositLocalContentCredentialsProvider:
     """`CredentialsProvider` implementation which provides a fallback for local development using a client credentials flow.
 
-    There is an open issue against the Databricks CLI which prevents it from returning service principal access tokens. 
+    There is an open issue against the Databricks CLI which prevents it from returning service principal access tokens.
     https://github.com/databricks/cli/issues/1939
 
     Until the CLI issue is resolved, this CredentialsProvider implements the approach described in the Databricks documentation
-    for manually generating a workspace-level access token using OAuth M2M authentication. Once it has acquired an access token, 
+    for manually generating a workspace-level access token using OAuth M2M authentication. Once it has acquired an access token,
     it returns it as a Bearer authorization header like other `CredentialsProvider` implementations.
 
     See Also
@@ -103,7 +104,7 @@ class PositLocalContentCredentialsProvider:
 
     def __call__(self) -> Dict[str, str]:
         response = requests.post(
-            self._token_endpoint_url, 
+            self._token_endpoint_url,
             auth=(self._client_id, self._client_secret),
             data={
                 "grant_type": "client_credentials",
@@ -114,6 +115,7 @@ class PositLocalContentCredentialsProvider:
 
         credentials = Credentials(**response.json())
         return _new_bearer_authorization_header(credentials)
+
 
 class PositContentCredentialsProvider:
     """`CredentialsProvider` implementation which initiates a credential exchange using a content-session-token.
@@ -136,10 +138,10 @@ class PositContentCredentialsProvider:
 
 class PositCredentialsProvider:
     """`CredentialsProvider` implementation which initiates a credential exchange using a user-session-token.
-    
-    The user-session-token is provided by Connect through the HTTP session header 
+
+    The user-session-token is provided by Connect through the HTTP session header
     `Posit-Connect-User-Session-Token`.
-    
+
     See Also
     --------
     * https://github.com/posit-dev/posit-sdk-py/blob/main/src/posit/connect/oauth/oauth.py
@@ -154,10 +156,11 @@ class PositCredentialsProvider:
         credentials = self._client.oauth.get_credentials(self._user_session_token)
         return _new_bearer_authorization_header(credentials)
 
+
 class PositLocalContentCredentialsStrategy(CredentialsStrategy):
     """`CredentialsStrategy` implementation which supports local development using OAuth M2M authentication against Databricks.
 
-    There is an open issue against the Databricks CLI which prevents it from returning service principal access tokens. 
+    There is an open issue against the Databricks CLI which prevents it from returning service principal access tokens.
     https://github.com/databricks/cli/issues/1939
 
     Until the CLI issue is resolved, this CredentialsStrategy provides a drop-in replacement as a local_strategy that can be used
@@ -165,17 +168,20 @@ class PositLocalContentCredentialsStrategy(CredentialsStrategy):
 
     Examples
     --------
-    In the example below, the `PositContentCredentialsStrategy` can be initialized anywhere that 
+    In the example below, the `PositContentCredentialsStrategy` can be initialized anywhere that
     the Python process can read environment variables.
 
     CLIENT_ID and CLIENT_SECRET are credentials associated with the Databricks service principal.
 
     ```python
-    from posit.connect.external.databricks import PositContentCredentialsStrategy, PositLocalContentCredentialsStrategy
+    from posit.connect.external.databricks import (
+        PositContentCredentialsStrategy,
+        PositLocalContentCredentialsStrategy,
+    )
 
     import pandas as pd
     from databricks import sql
-    from databricks.sdk.core import ApiClient, Config 
+    from databricks.sdk.core import ApiClient, Config
     from databricks.sdk.service.iam import CurrentUserAPI
 
     DATABRICKS_HOST = "<REDACTED>"
@@ -183,18 +189,18 @@ class PositLocalContentCredentialsStrategy(CredentialsStrategy):
     SQL_HTTP_PATH = "<REDACTED>"
     TOKEN_ENDPOINT_URL = f"https://{DATABRICKS_HOST}/oidc/v1/token"
 
-    CLIENT_ID = "<REDACTED>" 
-    CLIENT_SECRET = "<REDACTED>" 
+    CLIENT_ID = "<REDACTED>"
+    CLIENT_SECRET = "<REDACTED>"
 
-    # Rather than relying on the Databricks CLI as a local strategy, we use 
+    # Rather than relying on the Databricks CLI as a local strategy, we use
     # PositLocalContentCredentialsStrategy as a drop-in replacement.
-    # Can be replaced with the Databricks CLI implementation when 
+    # Can be replaced with the Databricks CLI implementation when
     # https://github.com/databricks/cli/issues/1939 is resolved.
     local_strategy = PositLocalContentCredentialsStrategy(
         token_endpoint_url=TOKEN_ENDPOINT_URL,
         client_id=CLIENT_ID,
         client_secret=CLIENT_SECRET,
-    ) 
+    )
 
     posit_strategy = PositContentCredentialsStrategy(local_strategy=local_strategy)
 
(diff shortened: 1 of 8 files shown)


zackverham:
Here's a code snippet that illustrates the ergonomics of using this code in published content (this is pulled from a Quarto doc I have running locally): 

```
from posit.connect.external.databricks import PositContentCredentialsStrategy

import requests

import pandas as pd
from databricks import sql

# env vars
DATABRICKS_HOST = "<REDACTED>"
DATABRICKS_HOST_URL = f"https://{DATABRICKS_HOST}"
SQL_HTTP_PATH = "<REDACTED>"

posit_strategy = PositContentCredentialsStrategy(local_strategy=databricks_cli)

cfg = Config(host=DATABRICKS_HOST_URL, credentials_strategy=posit_strategy)

databricks_user_info = CurrentUserAPI(ApiClient(cfg)).me()
print(f"Hello, {databricks_user_info.display_name}!")

query = "SELECT * FROM samples.nyctaxi.trips LIMIT 10;"
with sql.connect(
    server_hostname=DATABRICKS_HOST,
    http_path=SQL_HTTP_PATH,
    credentials_provider=posit_strategy.sql_credentials_provider(cfg),
) as connection:
    with connection.cursor() as cursor:
        cursor.execute(query)
        rows = cursor.fetchall()
        print(pd.DataFrame([row.asDict() for row in rows]))
```

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +31,80 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
-        access_token = credentials.get("access_token")
-        if access_token is None:
-            raise ValueError("Missing value for field 'access_token' in credentials.")
-        return {"Authorization": f"Bearer {access_token}"}
+        return _new_bearer_authorization_header(credentials)
 
 
-class PositCredentialsStrategy(CredentialsStrategy):
+class PositContentCredentialsStrategy(CredentialsStrategy):
+    """CredentialsStrategy implementation which returns a PositContentCredentialsProvider when called."""
```
schloerke, line 99: - [ ] Can you add your example quarto code here? (Or somewhere within the class. Maybe on the `__call__` method?)
    * This way if you hover over the object, you can see example usage. Thank you

- [ ] Can we also update the other classes with `Examples` sections as well? Thank you!

schloerke: Can we adjust the wording to expand on the local / hosted differences? (Doesn't have to be the suggestion below)

```suggestion
    """
    Posit content dynamic `CredentialsStrategy` implementation
    
    This strategy callable class returns a `PositContentCredentialsProvider` when hosted and its `local_strategy` strategy otherwise.  
    """
```

src/posit/connect/external/databricks.py
```diff
@@ -66,23 +119,46 @@ def sql_credentials_provider(self, *args, **kwargs):
...
-        otherwise, we should return the auth_type of the configured local_strategy instead
-        to avoid breaking someone elses attribution.
+        return PositContentCredentialsProvider(self._client)
 
-        https://github.com/databricks/databricks-sdk-py/blob/v0.29.0/databricks/sdk/config.py#L261-L269
 
-        NOTE: The databricks-sql client does not use auth_type to set the user-agent.
-        https://github.com/databricks/databricks-sql-python/blob/v3.3.0/src/databricks/sql/client.py#L214-L219
+class PositCredentialsStrategy(CredentialsStrategy):
+    """CredentialsStrategy implementation which returns a PositContentCredentialsProvider when called."""
```
schloerke, line 136: Please make a similar docs adjustment here

src/posit/connect/external/databricks.py
```diff
@@ -66,23 +119,46 @@ def sql_credentials_provider(self, *args, **kwargs):
...
+
+    def sql_credentials_provider(self, *args, **kwargs):
+        """The sql connector attempts to call the credentials provider w/o any args.
+
+        The SQL client's `ExternalAuthProvider` is not compatible w/ the SDK's implementation of
+        `CredentialsProvider`, so create a no-arg lambda that wraps the args defined by the real caller.
+        This way we can pass in a databricks `Config` object required by most of the SDK's `CredentialsProvider`
+        implementations from where `sql.connect` is called.
+
+        https://github.com/databricks/databricks-sql-python/issues/148#issuecomment-2271561365
```
schloerke, line 156:
```suggestion
        See Also
        --------
        * https://github.com/databricks/databricks-sql-python/issues/148#issuecomment-2271561365
```

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +31,80 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
+    The databricks-sdk client uses the configurated auth_type to create
+    a user-agent string which is used for attribution. We should only
+    overwrite the auth_type if we are using the PositCredentialsStrategy (non-local),
+    otherwise, we should return the auth_type of the configured local_strategy instead
+    to avoid breaking someone elses attribution.
+
+    https://github.com/databricks/databricks-sdk-py/blob/v0.29.0/databricks/sdk/config.py#L261-L269
+
+    NOTE: The databricks-sql client does not use auth_type to set the user-agent.
+    https://github.com/databricks/databricks-sql-python/blob/v3.3.0/src/databricks/sql/client.py#L214-L219
```
schloerke, line 62:
```suggestion
    NOTE: The databricks-sql client does not use auth_type to set the user-agent.
    https://github.com/databricks/databricks-sql-python/blob/v3.3.0/src/databricks/sql/client.py#L214-L219

    See Also
    --------
    * https://github.com/databricks/databricks-sdk-py/blob/v0.29.0/databricks/sdk/config.py#L261-L269
```

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +31,80 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
+    """
+    access_token = credentials.get("access_token")
+    if access_token is None:
+        raise ValueError("Missing value for field 'access_token' in credentials.")
+    return {"Authorization": f"Bearer {access_token}"}
+
+def _get_auth_type(local_auth_type: str) -> str: 
+    """Returns the auth type currently in use.
+
+    The databricks-sdk client uses the configurated auth_type to create
```
christierney, line 53: "configurated"?

zackverham: (it does appear configurated is a word, but this should almost certainly be configured) 

https://www.merriam-webster.com/dictionary/configurated

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +31,120 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
+    from databricks import sql
+    from databricks.sdk.core import ApiClient, Config, databricks_cli
+    from databricks.sdk.service.iam import CurrentUserAPI
+
+    # env vars
+    DATABRICKS_HOST = "<REDACTED>"
+    DATABRICKS_HOST_URL = f"https://{DATABRICKS_HOST}"
+    SQL_HTTP_PATH = "<REDACTED>"
+
+    posit_strategy = PositContentCredentialsStrategy(local_strategy=databricks_cli)
```
dbkegley, line 126: The other credentialsStrategy docstring uses a sample py-shiny app so that the user can (hopefully) see that the location that the initialization happens is really important. That context is lost in this example. I think it would be worthwhile to show both and call out explicitly why the initialization is different for service accounts than for viewer auth

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +31,120 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
+
+    Example 
+    -------
+    from posit.connect.external.databricks import PositContentCredentialsStrategy
+
+    import pandas as pd
+    import requests
+    
+    from databricks import sql
+    from databricks.sdk.core import ApiClient, Config, databricks_cli
```
dbkegley, line 118: @zackverham did you ever figure out a workaround for getting local service accounts tokens via the cli fallback?

src/posit/connect/external/databricks.py
zackverham: yeah, I guess the question ultimately is how much of this documentation should exist in the docstrings vs in our docs pages (both the cookbook examples themselves, and prose discussions explaining things). 

I think this might be a question for @schloerke - I don't know exactly what expectations would be for SDK consumers.

src/posit/connect/external/databricks.py
zackverham: I have not. Databricks labeled my ticket in the CLI repo as a bug, but that's where I'm stuck currently: 

https://github.com/databricks/cli/issues/1939

I'm not totally sure what we want to do with that. Options seem to be: 

1. remove the local option from our implementation and re-introduce it when databricks supports it
2. leave the option to provide a local strategy + code comments explaining the databricks cli doesn't support it. A particularly motivated user could probably fork the databricks cli and implement a token command that follows the pattern established in other cli operations to get it to work. This isn't super likely and obviously pretty rough, but the local strategy here doesn't explicitly state it has to be provided by an official vendor.

src/posit/connect/external/databricks.py
schloerke: I think docs strings should show usage examples. 

I think the current Connect recipes are turning into small usage examples and should be moved to the doc strings. (And the recipes should be larger/longer goal-oriented stories combining many small usage examples.)

-----------------

tl/dr; More usage examples in doc strings, please!

schloerke (approved):

schloerke:
Happy to have more usage examples if you'd like. Either way, feel free to merge when ready.

christierney (approved):

src/posit/connect/external/databricks.py
mconflitti-pbc: This example needs to be updated to use the `PositLocalContentCredentialsStrategy`

src/posit/connect/external/databricks.py
```diff
@@ -66,23 +306,107 @@ def sql_credentials_provider(self, *args, **kwargs):
...
-        https://github.com/databricks/databricks-sql-python/blob/v3.3.0/src/databricks/sql/client.py#L214-L219
+    app_ui = ui.page_fluid(ui.output_text("text"), ui.output_data_frame("result"))
+
+
+    def server(i: Inputs, o: Outputs, session: Session):
+
+        # HTTP session headers are available in this context.
+        session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
+        posit_strategy = PositCredentialsStrategy(
+            local_strategy=databricks_cli, user_session_token=session_token
```
mconflitti-pbc, line 355: should this now be the patched local strategy?

zackverham: its sort of ambiguous imo. the user could still use the databricks_cli as the local strategy, if say for example their databricks admin doesn't want to hand out service principal credentials. They can still use their own identity for local development with the databricks cli - that was my thinking for keeping this as-is.

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +39,259 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
+    There is an open issue against the Databricks CLI which prevents it from returning service principal access tokens. 
+    https://github.com/databricks/cli/issues/1939
+
+    Until the CLI issue is resolved, this CredentialsProvider implements the approach described in the Databricks documentation
+    for manually generating a workspace-level access token using OAuth M2M authentication. Once it has acquired an access token, 
+    it returns it as a Bearer authorization header like other `CredentialsProvider` implementations.
+
+    See Also
+    --------
+    * https://docs.databricks.com/en/dev-tools/auth/oauth-m2m.html#manually-generate-a-workspace-level-access-token
```
mconflitti-pbc, line 96: Seeing this example as well: https://docs.databricks.com/en/dev-tools/python-sql-connector.html#oauth-machine-to-machine-m2m-authentication

Is the problem using their version of this that they have omitted `client_credentials` scope?
https://github.com/databricks/databricks-sdk-py/blob/980166030c688e08a52d0f55389abd766f1d928d/databricks/sdk/credentials_provider.py#L164-L174

src/posit/connect/external/databricks.py
```diff
@@ -29,29 +39,259 @@ def __call__(self, *args, **kwargs) -> CredentialsProvider:
...
+
+    # Rather than relying on the Databricks CLI as a local strategy, we use 
+    # PositLocalContentCredentialsStragtegy as a drop-in replacement.
+    # Can be replaced with the Databricks CLI implementation when 
+    # https://github.com/databricks/cli/issues/1939 is resolved.
+    local_strategy = PositLocalContentCredentialsStrategy(
+        token_endpoint_url=TOKEN_ENDPOINT_URL,
+        client_id=CLIENT_ID,
+        client_secret=CLIENT_SECRET,
+    )
```
mconflitti-pbc, line 197: will it be on the user to make sure they have the proper env vars wired up? Curious if it makes sense to pull the values from their .databrickscfg file for them?

https://docs.databricks.com/en/dev-tools/cli/authentication.html#oauth-machine-to-machine-m2m-authentication

src/posit/connect/external/databricks.py
zackverham: Added some comments to the docstring to clarify the various use cases.

src/posit/connect/external/databricks.py
mconflitti-pbc: Maybe just a nice to have for now.

src/posit/connect/external/databricks.py
mconflitti-pbc: ah good point!

src/posit/connect/external/databricks.py
zackverham: its a good thought, for sure - I'm hesitant to add that, since (a) this is really just supposed to be a band-aid to get people unstuck until the databricks CLI is working again, and (b) we'd then have to implement a parser that understands the databrickscfg file (profile settings, etc).

src/posit/connect/external/databricks.py
zackverham: Good question - this is somewhat confusing, because the development of their CLI/SDK tooling and the sql connector are happening on independent threads and don't always sync up. 

Here's the credentials provider we care about - it should fall back to whatever the CLI does to acquire a token:
https://github.com/databricks/databricks-sdk-py/blob/980166030c688e08a52d0f55389abd766f1d928d/databricks/sdk/credentials_provider.py#L630

Which calls this: 
https://github.com/databricks/databricks-sdk-py/blob/980166030c688e08a52d0f55389abd766f1d928d/databricks/sdk/credentials_provider.py#L578

Which executes the `databricks auth token` command to acquire a token. 

The real crux of the problem is in the `auth token` command - their token command isn't written to support anything aside from auth code flows: 

https://github.com/databricks/cli/blob/main/cmd/auth/token.go#L45

So you can test this even outside of our code - just running `databricks auth token` with an oauth-m2m profile doesn't work. It ends up returning your own token and ignoring the profile.

mconflitti-pbc: Thanks for clarifying!

mconflitti-pbc (approved):

schloerke (approved):

PR #352, opened 2024-12-09
feat: Add support for system cache runtime

### LINKED ISSUE #337: Add System Runtime Caches endpoint bindings
Add support for the System Runtime Caches API. See https://docs.posit.co/connect/api/#get-/v1/system/caches/runtime

This should be implemented as a new resource and resemble the following pseudo-code:

```py
from posit import connect
from posit.connect.system import System, SystemCache

client = connect.Client()
system: System = client.system
caches: List[SystemCache]: system.caches
for cache in caches:
    task = cache.destroy(dry_run=True)
```

schloerke:
Fixes #337 

Example usage:

```python

client = Client()

runtime_caches = client.system.caches.runtime.find()
first_runtime_cache = runtime_caches[0]

# Remove the cache
task = first_runtime_cache.destroy(dry_run=True)

# Wait for the task to finish
task.wait_for()
```

### DIFF
--- a/integration/tests/posit/connect/test_system.py
+++ b/integration/tests/posit/connect/test_system.py
@@ -1,4 +1,3 @@
-import time
 from pathlib import Path
 
 import pytest
--- a/integration/tests/posit/connect/test_system.py
+++ b/integration/tests/posit/connect/test_system.py
@@ -52,6 +52,10 @@ class TestSystem:
                 raise TimeoutError("Timeout while waiting for cache to deploy")
 
             jobs = self.content_item.jobs.fetch()
+            if len(list(jobs)) == 0:
+                # No jobs found, restart while loop
+                continue
+
             for job in jobs:
                 if str(job["status"]) == "0":
                     # Stop for loop and restart while loop
--- a/integration/tests/posit/connect/test_system.py
+++ b/integration/tests/posit/connect/test_system.py
@@ -74,4 +74,6 @@ class TestSystem:
         # Check if the cache is deployed
         caches = self.client.system.caches.runtime.find()
 
-        assert len(caches) > 0
+        # Caches only added in Connect versions >= 2024.05.0
+        if CONNECT_VERSION >= version.parse("2024.05.0"):
+            assert len(caches) > 0
--- a/integration/tests/posit/connect/test_system.py
+++ b/integration/tests/posit/connect/test_system.py
@@ -84,7 +84,10 @@ class TestSystem:
         # Check if the cache is deployed
         caches = self.client.system.caches.runtime.find()
 
-        # Caches only added in Connect versions >= 2024.05.0
+        # Barret 2024/12:
+        # Caches only showing up in Connect versions >= 2024.05.0
+        # I do not know why.
+        # Since we are not logic testing Connect, we can confirm our code works given more recent versions of Connect.
         if CONNECT_VERSION >= version.parse("2024.05.0"):
             assert len(caches) > 0
 
--- a/integration/tests/posit/connect/test_system.py
+++ b/integration/tests/posit/connect/test_system.py
@@ -22,10 +22,7 @@ class TestSystem:
     def setup_class(cls):
         cls.client = Client()
         assert cls.client.content.count() == 0
-        cls.content_item = cls.client.content.create(
-            name="Content_A",
-            default_r_environment_management=True,
-        )
+        cls.content_item = cls.client.content.create(name="Content_A")
 
     # Copied from from integration/tests/posit/connect/test_packages.py
     def _deploy_python_bundle(self):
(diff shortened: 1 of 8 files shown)


schloerke:
- [ ] Dry run should return `None`

src/posit/connect/system.py
```diff
@@ -0,0 +1,292 @@
...
+
+from typing import TYPE_CHECKING, List, Literal, Optional, overload
+
+from typing_extensions import TypedDict, Unpack
+
+from .context import ContextManager
+from .resources import Active
+
+if TYPE_CHECKING:
+    from posit.connect.context import Context
```
tdstein, line 13:
```suggestion
    from .context import Context
```

src/posit/connect/system.py
```diff
@@ -0,0 +1,292 @@
...
+        # Remove the cache
+        task = first_runtime_cache.destroy(dry_run=False)
+
+        # Wait for the task to finish
+        task.wait_for()
+        ```
+        """
+        url = self._ctx.url + self._path
+        data = dict(self)
+        response = self._ctx.session.delete(url, json={**data, **kwargs})
```
tdstein, line 146: fwiw, I think this will also work...

```suggestion
        body = {**self, **kwargs}
        response = self._ctx.client.delete(self._path, json=body)
```

schloerke: It doesn't need the full url?

src/posit/connect/system.py
schloerke: Resolved with all other similarly shaped imports

src/posit/connect/system.py
tdstein: The client methods do the path append

schloerke: I missed the switch from `.session` to `.client`. 👍

integration/tests/posit/connect/test_system.py
```diff
@@ -0,0 +1,92 @@
...
+        caches: list[SystemRuntimeCache] = self.client.system.caches.runtime.find()
+        assert isinstance(caches, list)
+
+        # Remove all caches
+        self._remove_all_caches()
+
+        # Deploy a new cache
+        self._deploy_python_bundle()
+
+        # Wait for the cache jobs to finish
```
tdstein, line 81: Just curious. How is the Job representing the cache different than the deployment Task?

src/posit/connect/system.py
```diff
@@ -0,0 +1,291 @@
...
+        client = Client()
+
+        runtime_caches = client.system.caches.runtime.find()
+        ```
+        """
+        url = self._ctx.url + self._path
+        response = self._ctx.session.get(url)
+        results = response.json()
+        if not isinstance(results, dict) and "caches" not in results:
+            raise RuntimeError("`caches=` not found in response.")
```
tdstein, line 196: Is there a known scenario where `caches` doesn't exist? On a fresh server? Otherwise, this check isn't necessary. This would indicate that something else went wrong, which should result in a server error.

src/posit/connect/system.py
```diff
@@ -0,0 +1,291 @@
...
+        ```python
+        from posit.connect import Client
+
+        client = Client()
+
+        runtime_caches = client.system.caches.runtime.find()
+        first_runtime_cache = runtime_caches[0]
+
+        # Remove the cache
+        task = client.system.caches.runtime.destroy(first_runtime_cache, dry_run=False)
```
tdstein, line 270: Should the user just run `first_runtime_cache.destroy()` instead? Not sure if this method is adding additional benefit to the API.

src/posit/connect/system.py
tdstein: Either one is fine for now. But we could remove _ctx.session in a refactor.

schloerke: I like it a bunch. Making a separate PR that does more of this work.

tdstein: Great. It's a good debt-maintenance task if you have time.

schloerke: Done in #360 

Resolving

src/posit/connect/system.py
schloerke: Agreed. Removing!

tdstein (approved):
Nicely done! I would like to understand why the cache isn't being created when deploying content earlier than 2024.05.0, but that shouldn't block this change.

integration/tests/posit/connect/test_system.py
```diff
@@ -0,0 +1,92 @@
...
+        self._wait_for_cache_jobs()
+
+        # Check if the cache is deployed
+        caches = self.client.system.caches.runtime.find()
+
+        # Caches only added in Connect versions >= 2024.05.0
+        if CONNECT_VERSION >= version.parse("2024.05.0"):
```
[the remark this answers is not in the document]
schloerke:
```
        # Barret 2024/12:
        # Caches only showing up in Connect versions >= 2024.05.0
        # I do not know why.
        # Since we are not logic testing Connect, we can confirm our code works given more recent versions of Connect.
```

PR #355, opened 2024-12-12
feat: adds environments

tdstein:


src/posit/connect/resources.py
```diff
@@ -234,3 +234,83 @@ def find_by(self, **conditions: Any) -> T | None:
...
+    def destroy(self) -> None:
+        self._ctx.client.delete(self._path)
+
+    def update(self, **attributes):  # type: ignore
+        response = self._ctx.client.put(self._path, json=attributes)
+        result = response.json()
+        super().update(**result)
+
+
+class _ResourceSequence(Sequence):
```
schloerke, line 254: Should this class inherit from `Sequence[T]` and we replace the `Any` types below with `T`?

This might replace the need for a `_GetItem` class : https://github.com/posit-dev/posit-sdk-py/pull/356#discussion_r1882797688

src/posit/connect/resources.py
```diff
@@ -234,3 +234,83 @@ def find_by(self, **conditions: Any) -> T | None:
...
+class _Resource(dict):
+    def __init__(self, ctx: Context, path: str, **attributes):
+        self._ctx = ctx
+        self._path = path
+        super().__init__(**attributes)
+
+    def destroy(self) -> None:
+        self._ctx.client.delete(self._path)
+
+    def update(self, **attributes):  # type: ignore
```
schloerke, line 248: Can we replace it with the pyright ignore command? `# pyright: ignore[RULE]`
(And if it's only parameter specific, can we put it on an isolated line only? Thank you!)

schloerke (approved):

src/posit/connect/resources.py
tdstein: 👍🏻 I'll do that in a follow up.

PR #356, opened 2024-12-12
refactor: use static duck typing for packages

tdstein:


integration/Makefile
```diff
@@ -140,5 +140,5 @@ test:
 	set -o pipefail; \
 		CONNECT_VERSION=${CONNECT_VERSION} \
 		CONNECT_API_KEY="$(shell $(UV) run rsconnect bootstrap -i -s http://connect:3939 --raw)" \
-		$(UV) run pytest -s --junit-xml=./reports/$(CONNECT_VERSION).xml | \
+		$(UV) run pytest -s -k TestPackages --junit-xml=./reports/$(CONNECT_VERSION).xml | \
```
schloerke, line 143:
```suggestion
		$(UV) run pytest -s --junit-xml=./reports/$(CONNECT_VERSION).xml | \
```

src/posit/connect/packages.py
```diff
@@ -1,213 +1,99 @@
...
-        """Package description hash for R packages."""
+class _Package(Mapping[str, Any]):
+    pass
 
-        bundle_id: Required[str]
-        """The unique identifier of the bundle this package is associated with"""
 
-        app_id: Required[str]
-        """The numerical identifier of the application this package is associated with"""
+class _Packages(Sized, Protocol):
```
schloerke, line 54: Is it possible to suffix the class names with `P` to help visually declare that it's a protocol class?

```suggestion
class _PackagesP(Sized, Protocol):
```

(Much easier to perform with "rename symbol")

src/posit/connect/packages.py
```diff
@@ -1,213 +1,99 @@
...
+    def __getitem__(self, index: SupportsIndex) -> _ContentPackage: ...
 
-class ContentPackage(Active):
-    class _Package(TypedDict):
-        language: Required[str]
-        name: Required[str]
-        version: Required[str]
-        hash: Required[Optional[str]]
+    @overload
+    def __getitem__(self, index: slice) -> _ContentPackage: ...
```
schloerke, line 16:
```suggestion
    def __getitem__(self, index: slice) -> list[_ContentPackage]: ...
```

schloerke, line 16: Is it possible to put this into a generic helper class?

```python
class _GetItem(Sized, Generic[T], Protocol):
    @overload
    def __getitem__(self, index: SupportsIndex) -> T: ...

    @overload
    def __getitem__(self, index: slice) -> list[T]: ...
```

schloerke (approved):

integration/Makefile
tdstein: Whoops. Good catch!

tdstein: Added a change to alleviate the need to manually changing this command as part of the development workflow https://github.com/posit-dev/posit-sdk-py/pull/358

src/posit/connect/packages.py
tdstein: Actually, I think we should make these public types. External users may want to use them for type hinting.

For example...

```py
from posit.connect import Client, Packages

client = Client()
packages: Packages = client.packages
```

PR #357, opened 2024-12-12
refactor: use static duck typing for jobs and packages

tdstein:


schloerke (approved):
So much deletion. This is awesome!

PR #360, opened 2024-12-13
refactor: Leverage Context's Client to make requests

schloerke:
Original request: https://github.com/posit-dev/posit-sdk-py/pull/352#pullrequestreview-2502579569

I like it. So I updated as many as I could where I didn't have to guess at the intent. `Sessions` is still used by other classes that use `.params`.

Given all `Paginator` instances could use context, I switched those as well.


tdstein:
> Given all Paginator instances could use context, I switched those as well.

Thank you!

tdstein (approved):
🙌🏻 🙌🏻 🙌🏻  Nice net deletion!

PR #364, opened 2024-12-16
feat: add base protocols for Resource and ResourceSequence

tdstein:


src/posit/connect/resources.py
```diff
@@ -77,7 +88,26 @@ def update(self, **attributes):  # type: ignore[reportIncompatibleMethodOverride
         super().update(**result)
 
 
-class _ResourceSequence(Sequence):
+T = TypeVar("T", bound=Resource)
+
+
+class ResourceSequence(Protocol[T]):
```
schloerke, line 94: Thank you

schloerke (approved):

src/posit/connect/resources.py
```diff
@@ -62,7 +69,11 @@ def __init__(self, ctx: Context, path: str, /, **attributes):
         self._path = path
 
 
-class _Resource(dict):
+class Resource(Protocol):
+    def __getitem__(self, key: Hashable) -> Any: ...
+
+
+class _Resource(dict, Resource):
```
schloerke, line 76: This class is no longer needed?

src/posit/connect/resources.py
```diff
@@ -62,7 +69,11 @@ def __init__(self, ctx: Context, path: str, /, **attributes):
         self._path = path
 
 
-class _Resource(dict):
+class Resource(Protocol):
+    def __getitem__(self, key: Hashable) -> Any: ...
```
schloerke, line 73:
```suggestion
    def __getitem__(self, key: Hashable, /) -> Any: ...
```

src/posit/connect/resources.py
tdstein: Correct. The BaseResource class should go away. But that likely requires finishing the migration to the new implicit subtyping pattern. Which I am planning to complete.

PR #366, opened 2024-12-16
refactor: Use new duck typing for content item `.repository` methods

schloerke:
Related: #364 

Other changes: 
- Remove `_api.py`, `_api_call.py`, `_json.py` files as they are unused
- Move to new file to avoid crowding
- Use "mixin" to keep all methods close to each other
- Changed `.update()` method to update in place
	- [Docs for connect](https://github.com/rstudio/connect/blob/06c767f4be1f7928781f24e14ff302b44b13b036/docs/cookbook/content/updating-content-git-repository-settings/index.qmd#L52-L60) **do not** need to be updated.


src/posit/connect/_repository.py
```diff
@@ -0,0 +1,145 @@
...
+
+# ContentItem Repository uses a PATCH method, not a PUT for updating.
+class _ContentItemRepository(_Resource):
+    def update(self, **attributes):
+        response = self._ctx.client.patch(self._path, json=attributes)
+        result = response.json()
+        # # Calling this method will call `_Resource.update` which will try to PUT to the path.
+        # super().update(**result)
+        # Instead, update the dict directly.
+        dict.update(self, **result)
```
schloerke, line 32: This might be controversial, but I couldn't find a way to skip `_Resource`'s `.update()` method and only perform the `dict`'s update. 

By passing in self to `dict.update(self, **result)`, I was able to have full control over what was being performed. 

Happy to use a different solution, but I couldn't come up with a clean approach.

src/posit/connect/_repository.py
```diff
@@ -0,0 +1,145 @@
...
+    overload,
+)
+
+from .errors import ClientError
+from .resources import Resource, _Resource
+
+
+# TODO-barret: Replace with Resource class from https://github.com/posit-dev/posit-sdk-py/pull/364/files#diff-94b7dc3c7d7d7c7b1a5f25e06c37df5fc53e1921cb10d41d4f04b18a715fae55R72
+class ResourceP(Protocol):
+    def __getitem__(self, key: Hashable, /) -> Any: ...
```
schloerke, line 21: TODO once #364 is merged

src/posit/connect/_repository.py
tdstein: I agree it's unusual, nevertheless it's correct.

Another option is... This may be easier to understand, but it isn't as simple. I don't have strong opinions either way. If you decide to keep the implementation above, then you could add the following snippet to give the reader a quicker understanding of what the method is doing.

```
        result: dict = response.json()
        for key, value in result.items():
            self[key] = value
```

integration/tests/posit/connect/test_content_item_repository.py
```diff
@@ -65,7 +67,7 @@ def test_create_get_update_delete(self):
         assert content_repo is not None
 
         def assert_repo(r: ContentItemRepository):
-            assert isinstance(content_repo, ContentItemRepository)
+            assert isinstance(content_repo, _Resource)
```
tdstein, line 70: This check should pass as `assert isinstance(content_repo, ContentItemRepository)`. Do you need to add `@runtime_checkable`?

src/posit/connect/_repository.py
```diff
@@ -0,0 +1,145 @@
...
+    Content items GitHub repository information.
+
+    See Also
+    --------
+    * Get info: https://docs.posit.co/connect/api/#get-/v1/content/-guid-/repository
+    * Delete info: https://docs.posit.co/connect/api/#delete-/v1/content/-guid-/repository
+    * Update info: https://docs.posit.co/connect/api/#patch-/v1/content/-guid-/repository
+    """
+
+    @abstractmethod
```
tdstein, line 46: The `abstractmethod` is no longer needed since this is a pure protocol now.

src/posit/connect/_repository.py
tdstein, line 1: nit; maybe drop the `_` for now to be consistent with the other files?

schloerke: Can do

src/posit/connect/_repository.py
```diff
@@ -0,0 +1,145 @@
...
+from .resources import Resource, _Resource
+
+
+# TODO-barret: Replace with Resource class from https://github.com/posit-dev/posit-sdk-py/pull/364/files#diff-94b7dc3c7d7d7c7b1a5f25e06c37df5fc53e1921cb10d41d4f04b18a715fae55R72
+class ResourceP(Protocol):
+    def __getitem__(self, key: Hashable, /) -> Any: ...
+
+
+# ContentItem Repository uses a PATCH method, not a PUT for updating.
+class _ContentItemRepository(_Resource):
```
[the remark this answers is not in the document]
schloerke: `_Resource` has a `.destroy()` method. That should cover it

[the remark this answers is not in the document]
tdstein: Thank you. Sorry, I thought I was looking at the interface.

integration/tests/posit/connect/test_content_item_repository.py
schloerke: Added

tdstein (approved):

tdstein:
Checks didn't kick off for some reason. It's probably related to the partial outage on GitHub.

PR #368, opened 2024-12-17
refactor: add proper return types to resource sequence methods

tdstein:


src/posit/connect/resources.py
```diff
@@ -129,32 +145,37 @@ def __str__(self) -> str:
...
 
-    def create(self, **attributes: Any) -> Any:
+    def create(self, **attributes: Any) -> _T:
         response = self._ctx.client.post(self._path, json=attributes)
         result = response.json()
         uid = result[self._uid]
         path = posixpath.join(self._path, uid)
-        return _Resource(self._ctx, path, **result)
+        resource = self._factory(self._ctx, path, **result)
+        resource = cast(_T, resource)
```
tdstein, line 154: I want to avoid this cast, but I haven't figured it out yet...

schloerke: I couldn't either... I tried to do this locally before you opened the PR. 😜

schloerke: From other repos, I've used an explicit type expectation before casting the value.  

```python
        resource: _T_co = self._factory(self._ctx, path, **result)
        resource = cast(_T, resource)
```

While it feels redundant, it gives me confidence to want to change `._factory()` and not have unexpected results.  This way, we know we are transferring from a `_T_co` to ` _T` (and not an `Any` to a `_T`).

src/posit/connect/resources.py
```diff
@@ -89,38 +93,50 @@ def update(self, **attributes):  # type: ignore[reportIncompatibleMethodOverride
...
 
 
-class _ResourceSequence(Sequence[T], ResourceSequence[T]):
-    def __init__(self, ctx: Context, path: str, *, uid: str = "guid"):
+class _ResourceSequence(Sequence[_T], ResourceSequence[_T]):
+    def __init__(
+        self,
+        ctx: Context,
+        path: str,
+        factory: ResourceFactory[_T_co] = _Resource,
```
schloerke, line 125: My preference is to put uncommon args that have defaults behind some sort of "name requirement" mechanism.

```suggestion
        *,
        factory: ResourceFactory[_T_co] = _Resource,
```

schloerke, line 125: We do not adjust `factory` or `uid`, but their default values are shared objects between instances: https://www.pythonmorsels.com/mutable-default-arguments/

(No changes necessary... just a thought to file away for a different debug day)

src/posit/connect/resources.py
```diff
@@ -89,38 +93,50 @@ def update(self, **attributes):  # type: ignore[reportIncompatibleMethodOverride
         super().update(**result)
 
 
-T = TypeVar("T", bound=Resource)
+_T = TypeVar("_T", bound=Resource)
+_T_co = TypeVar("_T_co", bound=Resource, covariant=True)
```
schloerke, line 97: A perfect use case!

Linking for my own reminder: https://stackoverflow.com/a/61737894/591574

src/posit/connect/resources.py
```diff
@@ -173,15 +194,16 @@ def find_by(self, **conditions) -> Any | None:
...
         paginator = Paginator(self._ctx, self._path, dict(**conditions))
         for page in paginator.fetch_pages():
             resources = []
             results = page.results
             for result in results:
                 uid = result[self._uid]
                 path = posixpath.join(self._path, uid)
-                resource = _Resource(self._ctx, path, **result)
+                resource = self._factory(self._ctx, path, **result)
+                resource = cast(_T, resource)
```
schloerke, line 207:
```suggestion
                resource: _T_co = self._factory(self._ctx, path, **result)
                resource = cast(_T, resource)
```

src/posit/connect/resources.py
schloerke, line 154: From other repos, I've used an explicit type expectation before casting the value.  

```python
        resource: _T_co = self._factory(self._ctx, path, **result)
        resource = cast(_T, resource)
```

While it feels redundant, it gives me confidence if we ever change `._factory()` and its return type.  This way, we know we are transferring from a `_T_co` to ` _T` (and not an `Any` to a `_T`).

src/posit/connect/resources.py
```diff
@@ -129,32 +145,37 @@ def __str__(self) -> str:
...
         return resources
 
-    def find(self, *args: str) -> Any:
+    def find(self, *args: str) -> _T:
         path = posixpath.join(self._path, *args)
         response = self._ctx.client.get(path)
         result = response.json()
-        return _Resource(self._ctx, path, **result)
+        resource = self._factory(self._ctx, path, **result)
+        resource = cast(_T, resource)
```
schloerke, line 175:
```suggestion
        resource: _T_co = self._factory(self._ctx, path, **result)
        resource = cast(_T, resource)
```

src/posit/connect/resources.py
```diff
@@ -129,32 +145,37 @@ def __str__(self) -> str:
...
         response = self._ctx.client.get(self._path, params=conditions)
         results = response.json()
-        resources = []
+        resources: List[_T] = []
         for result in results:
             uid = result[self._uid]
             path = posixpath.join(self._path, uid)
-            resource = _Resource(self._ctx, path, **result)
+            resource = self._factory(self._ctx, path, **result)
+            resource = cast(_T, resource)
```
schloerke, line 165:
```suggestion
            resource: _T_co = self._factory(self._ctx, path, **result)
            resource = cast(_T, resource)
```

src/posit/connect/resources.py
schloerke: But is it necessary since we're `cast()`ing?

I think we might be able to get away with a invariant type.

```suggestion
_R = TypeVar("_R", bound=Resource)
```

(`R` for factory _R_ eturn value?)

schloerke (changes requested):
Couple typing style requests.

One question on the covariant type var.

PR #376, opened 2025-01-17
fix: Update with_user_session_token method to be more robust and allow local development

mconflitti-pbc:
Follow up from discussions:
- https://github.com/posit-dev/connect/issues/29529
- https://positpbc.slack.com/archives/C077N436Q00/p1737140086322299

On hold until we confirm final terminology: visitor vs viewer with @kellobri


src/posit/connect/client.py
```diff
@@ -174,16 +177,27 @@ def __init__(self, *args, **kwargs) -> None:
         self._ctx = Context(self)
 
     @requires("2025.01.0-dev")
-    def with_user_session_token(self, token: str) -> Client:
+    def with_user_session_token(
+        self, token: str, fallback_api_key_env_var: str = "CONNECT_API_KEY"
```
toph-allen, line 181: In `connectapi`'s implementation, I originally would just fall back to the API key that the client already has. I thought that'd work because I'm just using the existing constructor function (https://github.com/posit-dev/connectapi/pull/363).

But something felt kinda wrong and typing up this comment, I think I figured it out.

```r
token <- session$headers$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN

# Using the default key as a fallback key
owner_client <- connect() # uses default api key
client <- connect(token = token)

# Using the a custom fallback key
owner_client <- connect() # uses default api key
client <- connect(api_key = Sys.getenv("FALLBACK_API_KEY"), token = token)
```

The latter won't work on Connect because its API key env var won't exist on Connect. So it _does_ need to be a separate argument. I'll probably still ask the key to be provided directly, so it's symmetrical with how the regular api key arg is provided.

---

Perhaps in this argument, the fallback key should also just be a string, so that it matches the way that API keys can optionally be provided (relying on the user to get them from an environment variable). It makes the name nicer too, i.e. `fallback_api_key`.

toph-allen (commented):
This generally looks good to me. Just one question / thought, but it's a very soft suggestion. I'll leave it to @tdstein to give the actual approval though. :)

src/posit/connect/client.py
mconflitti-pbc: Thanks Toph! Maybe I am misunderstanding, but what would the suggestion to change here?

nealrichardson: I'm not convinced that we need this argument added. If you want to use something other than `CONNECT_API_KEY`, initialize the client with `token = os.getenv("YOUR_ENV_VAR")`, right?

mconflitti-pbc: `token` is for the user session token which a user could not set locally. The fallback is there for them to define an API key that they created themselves for testing locally.

I suppose though you are right that token could take in an API key and then since is_local() is true, use token as the api key.. hmm. Then it is a question of being implicit or explicit.

Also, I think that could be problematic since `token = os.getenv("YOUR_ENV_VAR")` would not be able to be deployed to connect. I tried to write the function in a way that the user can write the code once and change behavior with the presence of env vars instead of needing to change the code. Though, I may be misunderstanding what you are proposing.

tdstein: I think for now, it's better to fail fast or return `self` when running locally. 

When calling this method in my consumer code, I wrapped it in conditional logic based on the existence of the `posit_connect_user_session_token`.  It's simple enough for others to do the same for now.

mconflitti-pbc: I like the simplicity of that approach. Seems to be a tension between do it for the user versus they need to handle it. Would prefer to keep the function simpler and fail fast. Trying to think if there is a bad consequence to that. Only thing I can think of is that for other integrations, is_local is handled automatically for them.

tdstein: Returning `self` when running locally seems like a decent middle ground. The documentation should just be clear.

mconflitti-pbc: Possible scenarios to handle:
- App that uses publisher client and visitor client
- App that only uses visitor client

I think that returning self doesnt allow a user to test the first scenario which was the motivation for adding that fallback. If we want to punt on that. Totally fine.

tdstein: Ah, good point. Failing fast and forcing the user to handle the conditional case seems ok for now. This is straightforward to change in the future as the use cases become more clear.

mconflitti-pbc: Simplified the method.

tdstein (approved):
Thanks for working through the feedback. Looks good to merge!

src/posit/connect/client.py
```diff
@@ -195,14 +197,18 @@ def with_user_session_token(self, token: str) -> Client:
...
+        visitor_credentials = self.oauth.get_credentials(
             token, requested_token_type=API_KEY_TOKEN_TYPE
         )
-        viewer_api_key = viewer_credentials.get("access_token")
-        if viewer_api_key is None:
-            raise ValueError("Unable to retrieve viewer api key.")
 
-        return Client(url=self.cfg.url, api_key=viewer_api_key)
+        visitor_api_key = visitor_credentials.get("access_token", "")
+        if visitor_api_key == "":
```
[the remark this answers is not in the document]
mconflitti-pbc: Only if something really gets messed up on the connect api side. Very unlikley but wanted to provide a better error message than what may show up otherwise.

PR #379, opened 2025-01-27
chore: added more docs/examples to with_user_session_token

mconflitti-pbc:
Added a couple more examples to make it clearer to people using the SDK how this function works to generate a visitor client.

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -182,6 +182,11 @@ class Client(ContextManager):
         a user session token will not exist, which will cause this method to result in an error needing
         to be handled in your application.
 
+        Depending on the type of application you are building, the user session token is retrieved in
+        a variety of ways. For example, in Streamlit and Shiny applications, the token is stored in the
+        context or session object headers using the `Posit-Connect-User-Session-Token` key. For API
+        applications, the token is added to the request headers.
+
         Parameters
         ----------
         token : str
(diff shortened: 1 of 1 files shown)


tdstein (approved):
Very nice. Thank you!

src/posit/connect/client.py
```diff
@@ -192,10 +197,50 @@ def with_user_session_token(self, token: str) -> Client:
...
+        >>> def user_profile():
+        ...     # fetch the viewer's profile information
+        ...     return visitor_client().me
+
+        Example with fallback when header is missing in API request:
+        >>> from posit.connect import Client
+        >>> import requests
+        >>>
+        >>> def get_client(request):
+        ...     base_client = Client()
```
jonkeane, line 235: What credentials would this `base_client` be authenticated with?

tdstein: CONNECT_URL and CONNECT_API_KEY. I don't think it is necessary to restate that here since it's mentioned in the Client documentation.

src/posit/connect/client.py
```diff
@@ -192,10 +197,50 @@ def with_user_session_token(self, token: str) -> Client:
...
+        ------
+        ValueError
+            If the provided token is `None` or empty or if the exchange response is malformed.
+        ClientError
+            If the token exchange request with the Connect Server fails.
+
         Examples
         --------
         >>> from posit.connect import Client
         >>> client = Client().with_user_session_token("my-user-session-token")
```
schloerke, line 210: Can the examples be done with python code blocks?

```suggestion
        ```python
        client = Client().with_user_session_token("my-user-session-token")
        ```
```

Please apply to other examples below

tdstein: The numpydoc's style guide recommends `>>>` instead: https://numpydoc.readthedocs.io/en/latest/format.html#examples

src/posit/connect/client.py
tdstein: Ah, good point. Yes, it's better to raise the exception here instead of returning the original client.

src/posit/connect/client.py
mconflitti-pbc: Hmm looks like there is a mix in the codebase. Which is preferred?

src/posit/connect/client.py
```diff
@@ -192,10 +197,54 @@ def with_user_session_token(self, token: str) -> Client:
...
+        >>> from posit.connect import Client
+        >>> import requests
+        >>>
+        >>> def get_client(request):
+        ...     base_client = Client()
+        ...     token = request.headers.get("Posit-Connect-User-Session-Token")
+        ...     if token:
+        ...         return base_client.with_user_session_token(token)
+        ...     else:
+        ...         return None
```
jonkeane, line 240:
```suggestion
        ...     # It's important to make catch if the visitor's token could not be retrieved (for example, if this app allows unauthenticated access) and handle that in cases where a token is expected (see below)

        ...     token = request.headers.get("Posit-Connect-User-Session-Token")
        ...     if token:
        ...         return base_client.with_user_session_token(token)
        ...     else:
        ...         return None
```

I'm not sure this is the right place for it, but having some text that explains why / when this is important would be nice. It also makes me think: maybe we need to think about a nicer way to not require there to be two `if`/`else` or `try`/`catch` blocks for something like this. We don't have to perfect it here, but this to me is a sign we might rethink how we tell folks to do this and see if there's a cleaner way we can provide a nice experience for someone using this feature.

src/posit/connect/client.py
tdstein: Technically, it is the REPL style, but Quartodoc currently renders both successfully. So, your call 😄

src/posit/connect/client.py
tdstein: Yeah, letting the use cases drive the design a bit...

I think the ideal is escaping the "wither" pattern and forcing client creation using either the API Key or the Session Token. But that isn't possible at the moment.

jonkeane: _nods_ yeah, and it's not _just_ the API Key or Session token theoretically. In the oauth world it might be: visitor Oauthcreds, then service account creds, then something that catches and fails gracefully. We won't solve that entirely here, but that's the experience we should aim to provide if we can

mconflitti-pbc: agreed! will simplify the example for now. can circle back if/when the visitor api key feature gets traction and we continue to evolve it.

src/posit/connect/client.py
```diff
@@ -192,10 +197,56 @@ def with_user_session_token(self, token: str) -> Client:
...
+            return client.with_user_session_token(user_session_token)
+        @render.text
+        def user_profile():
+            # fetch the viewer's profile information
+            return visitor_client().me
+        ```
+
+        Example of when the visitor's token could not be retrieved (for
+        example, if this app allows unauthenticated access) and handle
+        that in cases where a token is expected.
```
jonkeane, line 236: 💯 this is great, thank you

PR #384, opened 2025-02-14
feat: implement posit workbench credentials strategy and make credentials strategy fallback options more explicit

### LINKED ISSUE #382: Implement `WorkbenchManagedCredentialsStrategy` for Databricks helpers
### Problem

The external databricks helpers aren't intuitive to use with workbench yet. The `local_strategy=databricks_cli` fallback option does not work because workbench can't do an external browser redirect flow. A working alternative is to use `pat_auth` instead, but that's not obvious because workbench managed credentials use an oauth token, not a PAT.  ([`pat_auth`](https://github.com/databricks/databricks-sdk-py/blob/6f094aafffc13fbe5c236e6a7735b1b239b1ab78/databricks/sdk/credentials_provider.py#L128-L136) just sets a static bearer token on the auth header so it works the same, but how is a user supposed to know that?)

### Proposal 

We should implement a `posit.connect.external.databricks.PositWorkbenchManagedCredentialsStrategy` to make it more obvious to users how they should use managed workbench credentials in a piece of content that is going to be deployed to Connect.

#### Option 1:

```python
session_token = headers.get("Posit-Connect-User-Session-Token")

cfg = Config("workbench")
workbench_strategy = PositWorkbenchManagedCredentialsStrategy(cfg)

posit_strategy = PositCredentialsStrategy(
    local_strategy=workbench_strategy,
    user_session_token=session_token,
)
cfg.credentials_strategy = posit_strategy
```

There's still some awkwardness here though. Using `local_strategy=workbench_strategy` would not work locally because the `"workbench"` managed databricks profile wouldn't exist in the config. Also the need to modify `cfg.credentials_strategy` feels bad.

#### Option 2: prefer a more explicit configuration

```python
session_token = headers.get("Posit-Connect-User-Session-Token")

posit_strategy = PositCredentialsStrategy(
    local_strategy=databricks_cli,
    workbench_strategy=PositWorkbenchManagedCredentialsStrategy(Config("workbench"))
    user_session_token=session_token,
)

cfg = Config(
    host=DATABRICKS_HOST_URL,
    credentials_strategy=posit_strategy,
)
...
```

This is better but the use of the Connect strategy is still non-obvious. How does the caller know if it will do a client credentials flow or a viewer auth flow?

#### Option 3: an even more explicit configuration

```python
posit_strategy = PositCredentialsStrategy(
    local_strategy=databricks_cli,
    workbench_strategy=PositWorkbenchManagedCredentialsStrategy(Config("workbench"))
    connect_strategy=PositViewerCredentialsStrategy(headers.get("Posit-Connect-User-Session-Token"))
)

cfg = Config(
    host=DATABRICKS_HOST_URL,
    credentials_strategy=posit_strategy,
)
...
```

dbkegley:
Closes #382

These changes make the separation between the different credential strategies more explicit and tries to be better about detecting the difference between local and workbench content.

It also combines the `Content` and `Viewer` strategy into a single new class: `ConnectStrategy`. If `user-session-token` is provided then we use the viewer implementation. If not then we should fall back to the content credentials (service account) implementation.

I also modified some of the naming in the hopes of the choice of which strategy to use more obvious to end users of these helpers. 

This example shows how to construct a Databricks SDK `Config` that is compatible with:
- Databricks CLI authentication for local development
- Workbench-managed Databricks Credentials in Posit Workbench
- Viewer OAuth integration authentication in Posit Connect

```python
import os

from databricks.sdk.core import ApiClient
from databricks.sdk.core.credentials_provider import databricks_cli
from shiny import reactive
from shiny.express import render, session

from posit.workbench.external.databricks import WorkbenchStrategy
from posit.connect.external.databricks import (
    ConnectStrategy,
    databricks_config,
)


@reactive.calc
def cfg():
    session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
    return databricks_config(
        posit_default_strategy=databricks_cli,
        posit_workbench_strategy=WorkbenchStrategy(),
        posit_connect_strategy=ConnectStrategy(user_session_token=session_token),
        host=os.getenv("DATABRICKS_HOST"),
    )


@render.text
def text():
    databricks_user_info = CurrentUserAPI(ApiClient(cfg())).me()
    return f"Hello, {databricks_user_info.display_name}!"
```

### Testing 

This needs QA in Connect and Workbench if possible. 

Workbench should test using the `WorkbenchStrategy()` with Workbench-managed credentials.

Connect should test with `ConnectStrategy()` with both a Viewer and a Service Account OAuth integration.

It would be nice to test the `azure_service_principal` and `oauth_service_principal` strategies as a replacement for the deprecated `PositLocalContentCredentialsProvider` helper that was removed.


src/posit/connect/external/databricks.py
```diff
@@ -7,24 +7,28 @@
...
 POSIT_OAUTH_INTEGRATION_AUTH_TYPE = "posit-oauth-integration"
 POSIT_LOCAL_CLIENT_CREDENTIALS_AUTH_TYPE = "posit-local-client-credentials"
-
-# The Databricks SDK CredentialsProvider == Databricks SQL HeaderFactory
-CredentialsProvider = Callable[[], Dict[str, str]]
+POSIT_WORKBENCH_AUTH_TYPE = "posit-workbench"
 
 
-class CredentialsStrategy(abc.ABC):
+class CredentialsStrategyWrapper(CredentialsStrategy):
```
dbkegley, line 31: Need to revisit this. I'd rather not depend on the databricks sdk but this violates the `TYPE_CHECKING` import above.

src/posit/connect/external/databricks.py
```diff
@@ -395,42 +345,59 @@ def text():
...
 
     def __call__(self, *args, **kwargs) -> CredentialsProvider:
-        # If the content is not running on Connect then fall back to local_strategy
-        if is_local():
+        if is_connect() and self._connect_strategy is not None:
+            self._auth_type = self._connect_strategy.auth_type()
+            return self._connect_strategy(*args, **kwargs)
+        else:
+            print()
+            # log and continue
```
dbkegley, line 389: I'd like move these log messages into the init so that they aren't printed everytime the user requests credentials

src/posit/connect/external/databricks.py
```diff
@@ -234,105 +225,64 @@ def __init__(self, token_endpoint_url: str, client_id: str, client_secret: str):
...
 
     def auth_type(self) -> str:
         return POSIT_LOCAL_CLIENT_CREDENTIALS_AUTH_TYPE
 
     def __call__(self, *args, **kwargs) -> CredentialsProvider:  # noqa: ARG002
-        return PositLocalContentCredentialsProvider(
+        return positLocalContentCredentialsProvider(
             self._token_endpoint_url,
             self._client_id,
             self._client_secret,
```
mconflitti-pbc, line 236: Can you remind me again why we need this? Shouldn't we be able to just pass the user session token to client.oauth.get_credentials()?

src/posit/connect/external/databricks.py
```diff
@@ -234,105 +225,64 @@ def __init__(self, token_endpoint_url: str, client_id: str, client_secret: str):
...
-
-    def __init__(
-        self,
-        local_strategy: CredentialsStrategy,
-        client: Optional[Client] = None,
+    """
+    def __init__(self,
+                 client: Optional[Client] = None,
+                 user_session_token: Optional[str] = None,
     ):
```
mconflitti-pbc, line 264:
```suggestion
    """
    def __init__(self,
                 client: Optional[Client] = None,
                 *,
                 user_session_token: Optional[str] = None,
                 content_session_token: Optional[str] = None,
    ):
```

Would keep these explicitly keyword arguments and then allows the dev to pass the token with an appropriately named param.

kmasiello:
My review is not of the code itself, but from the perspective of the developer/publisher trying to understand the correct usage of this helper. Some of my comments may reflect naivety in Python, but this may be representative of many of our users and we want them to be able to use this helper without significant mental overhead.  

Here's a markup of the questions and confusion I had in trying to understand the shiny example.  I will summarize each main point (numbered) separately below for discussion.  (and no, i don't usually do PR comments with images 😂)
image

kmasiello:
(1). "Posit", "Credentials", "Strategy", "posit_strategy", "credentials_strategy", "connect_strategy", "PositCredentials", "PositConnectCredentials", "PositWorkbenchCredentials" ...
This is very distracting and overwhelming, making it difficult to follow the logic path for what we're doing here.

kmasiello:
(2). `session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")` or `session_token = flask.request.headers.get("Posit-Connect-User-Session-Token")`.  The viewer's session token is always going to be retrieved at this header, so don't make me have to write the code to go get it.

kmasiello:
(3). `workbench_strategy=PositWorkbenchCredentialsStrategy(Config(profile="workbench")),` I don't love this because it's a lot to type out.  Why not just `workbench_strategy=Config(profile="workbench")` or set this by default so I can just do `workbench_strategy=PositWorkbenchCredentialsStrategy()` but if I happen to have a different profile defined (maybe because I'm using a service account and need M2M) then I can change to a different profile.

kmasiello:
(4) `connect_strategy=PositConnectCredentialsStrategy(user_session_token=session_token)`. Again, this is a lot to type out.  and it would align better with my mental model of credential methods on connect if I could instead specify `connect_strategy=PositConnectCredentialsStrategy(type=[viewer_oath | service_account_oath | envvars])`

kmasiello:
(5). `PositWorkbenchCredentialsStrategy` and `PositConnectCredentialsStrategy` - again, word soup. Maybe we call these credential methods or credential types?  I think the helper should have one "Strategy", not a PositStrategy and a <Product>Strategy

kmasiello:
(6). `posit_strategy` isn't descriptive of what it actually is. It's a strategy for handling credentials, not for handling Posit.

kmasiello:
(7) here's where I completely lost the logic trail. 
Part of my confusion was not realizing that `credentials_provider` is a valid argument to `sql.connect`.  I had only used `access_token=` before. So among the word soup from posit_sdk and now the databricks sql connector adding a related term, it was hard to follow. I've stared at this for an hour and I still don't understand the flow here.

examples/connect/databricks/shiny/app.py
```diff
@@ -25,12 +29,11 @@ def server(i: Inputs, o: Outputs, session: Session):
     session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
```
kmasiello, line 29: A nit, and this applies to all the examples:
 From the developer’s POV, I am in local/workbench first. It feels out of order to have the Connect-specific session token defined so early. My mental model is to get through the parts about defining how to handle creds in development vs deployment, then I’d start putting Connect-specific info.

kmasiello:
🪱 🥫
Databricks SQL Connector is great. But what about authenticating to compute clusters using Spark and `databricks-connect` ?

examples/connect/databricks/dash/app.py
```diff
@@ -9,7 +9,11 @@
 from databricks.sdk.core import ApiClient, Config, databricks_cli
 from databricks.sdk.service.iam import CurrentUserAPI
 
-from posit.connect.external.databricks import PositCredentialsStrategy
+from posit.connect.external.databricks import (
+    PositCredentialsStrategy,
+    PositConnectCredentialsStrategy,
+    PositWorkbenchCredentialsStrategy,
```
nealrichardson, line 15: Just a passing observation: it's odd to see `PositWorkbenchCredentialsStrategy` inside the `posit.connect` module. They aren't really relevant for Connect, right? Is it time to create `posit.workbench` to contain these?

mconflitti-pbc:
> (2). `session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")` or `session_token = flask.request.headers.get("Posit-Connect-User-Session-Token")`. The viewer's session token is always going to be retrieved at this header, so don't make me have to write the code to go get it.

Great point! I think the trouble with doing this for people though is that its location is dependent on the app/framework they are using. Would be neat to automatically find it though, I agree.

src/posit/connect/external/databricks.py
dbkegley: Originally we had this because the databricks-cli [doesn't support service principal auth](https://github.com/databricks/cli/issues/1939) however after looking more closely at the databricks-sdk, I think we should be able to use some of the [provided service_principal credentials strategies](https://github.com/databricks/databricks-sdk-py/blob/998a117c43a7bc901710d263a7a7ab0d66ae8b8c/databricks/sdk/credentials_provider.py#L164-L185) rather than implementing this ourselves.

dbkegley:
> Databricks SQL Connector is great. But what about authenticating to compute clusters using Spark and databricks-connect ?

Any of the various `CredentialsStrategy` implementations from this PR or from the databricks-sdk should also work with `databricks-connect`

examples/connect/databricks/shiny/app.py
mconflitti-pbc: This could be a constraint given the custom nature of this integration. The way it is shown here, this code can be written once and then adapt to the environment it is running it. But that means there is a lowest common denominator in terms of DX as a tradeoff. Ideally, devs are thinking about how their code would work in production as well if that is where they plan to deploy to.

Referencing your other comments though, if we had a way to grab the user session token for the dev if/when it is needed then that could be done behind the scenes at the appropriate time simplifying this immensely.

mconflitti-pbc:
```python
posit_strategy = PositCredentialsStrategy(
        local_strategy=databricks_cli,
        user_session_token=session_token,
        workbench_strategy=PositWorkbenchCredentialsStrategy(Config(profile="workbench")),
        connect_strategy=PositConnectCredentialsStrategy(user_session_token=session_token),
    )
```
could become:
```python
posit_strategy = PositCredentialsStrategy()
```
allowing for overrides to defaults, but otherwise it just grabs what it needs under the hood.

dbkegley:
> (1). "Posit", "Credentials", "Strategy", "posit_strategy", "credentials_strategy", "connect_strategy", "PositCredentials", "PositConnectCredentials", "PositWorkbenchCredentials" ...
This is very distracting and overwhelming, making it difficult to follow the logic path for what we're doing here.

We can do my best to hide some of this from the user but this is an artifact of the databricks SDK, not something we are imposing in our client. 

https://github.com/databricks/databricks-sdk-py/blob/998a117c43a7bc901710d263a7a7ab0d66ae8b8c/databricks/sdk/config.py#L103-L121

> (2). `session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")` or `session_token = flask.request.headers.get("Posit-Connect-User-Session-Token")`. The viewer's session token is always going to be retrieved at this header, so don't make me have to write the code to go get it.

Right. As Matt said, this is framework-dependent. We could consider adding a helper for each framework but that feels like it will lead to even more confusion like we see with `(1)`

> (3). `workbench_strategy=PositWorkbenchCredentialsStrategy(Config(profile="workbench"))`, I don't love this because it's a lot to type out. Why not just `workbench_strategy=Config(profile="workbench")` or set this by default so I can just do `workbench_strategy=PositWorkbenchCredentialsStrategy()` but if I happen to have a different profile defined (maybe because I'm using a service account and need M2M) then I can change to a different profile.

Good call. I think we can do `workbench_strategy=PositWorkbenchCredentialsStrategy()` pretty easily.

> (4) `connect_strategy=PositConnectCredentialsStrategy(user_session_token=session_token)`. Again, this is a lot to type out. and it would align better with my mental model of credential methods on connect if I could instead specify `connect_strategy=PositConnectCredentialsStrategy(type=[viewer_oath | service_account_oath | envvars])`

We can revisit this but if the choice is `viewer_oauth` then _someone_ has to pass the token from the header into the `PositConnectCredentialsStrategy`.

The way this is implemented at the moment, if `user_session_token` is not provided then we attempt to default to Service Account auth, so the presence of `user_session_token` is what drives the type of auth used. This way if the publisher changes the oauth integration association in Connect from a Viewer integration to a Service Account integration, then they don't need to update any of their content code. If we want to make this an explicit choice in the code then we can definitely do that instead. I tend to prefer explicit options but this was one area where I tried to make the user do less typing.

> (5). `PositWorkbenchCredentialsStrategy` and `PositConnectCredentialsStrategy` - again, word soup. Maybe we call these credential methods or credential types? I think the helper should have one "Strategy", not a PositStrategy and a Strategy

If you want to write content that _only_ works on workbench then you don't need the `PositStrategy`, simply pass a `PositWorkbenchStrategy` when constructing the `Config(credentials_strategy=PositWorkbenchStrategy())`. The `PositStrategy` is useful when you want to author content that works in all 3 environments without making any code changes.

More broadly, the word choices "CredentialsStrategy" and "CredentialsProvider" are constructs from the databricks-sdk, not something we came up with here. We tried to be consistent with their naming in the SDK to avoid confusion but we can call these things whatever we want.

> (6). `posit_strategy` isn't descriptive of what it actually is. It's a strategy for handling credentials, not for handling Posit.

This one is just a var name so an easy fix. How about we just call this `creds` or `credentials`?

> (7) here's where I completely lost the logic trail.
Part of my confusion was not realizing that `credentials_provider` is a valid argument to `sql.connect`. I had only used `access_token=` before. So among the word soup from posit_sdk and now the databricks sql connector adding a related term, it was hard to follow. I've stared at this for an hour and I still don't understand the flow here.

This is a major part of the friction with implementing these helpers. We are trying to make something easy to do for our users when some of these libraries (databricks-sdk and databricks-sql) aren't even compatible inside of Databricks' own ecosystem of tools.

https://github.com/databricks/databricks-sql-python/issues/148#issuecomment-2271561365

dbkegley:
Oh an regarding the circular reference mentioned in `(7)`. Yes. It's another reason this is so challenging. `Config.credentials_strategy` is a `Callable` which is called with `Config` as an argument.

dbkegley:
@kmasiello could you take another look at this when you have a minute? 

I've done some refactoring based on your feedback. One of the main issues we had before was that there was this circular dependency between a `databricks.Config` and the `credentials_strategy`. I've tried to remove this by adding a `new_config` helper for constructing a databricks config that is all set up with the right strategy. The defaults should now work in _most_ environments but can be overridden explicitly if desired. I still need to test all the different combinations but the basic gist is:

```python
import streamlit as st
from databricks import sql
from databricks.sdk.core import ApiClient

from posit.connect.external.databricks import (
    new_config,
    sql_credentials,
    ConnectStrategy,
)

DATABRICKS_HOST = os.getenv("DATABRICKS_HOST")
DATABRICKS_HOST_URL = f"https://{DATABRICKS_HOST}"
SQL_HTTP_PATH = os.getenv("DATABRICKS_PATH")

session_token = st.context.headers.get("Posit-Connect-User-Session-Token")
cfg = new_config(
    posit_connect_strategy=ConnectStrategy(
        user_session_token=session_token
    ),
)

databricks_user = CurrentUserAPI(ApiClient(cfg)).me()
st.write(f"Hello, {databricks_user.display_name}!")
``` 

Unfortunately we can't really get rid of the need to pass in the session_token arg when specifying the connect strategy but hopefully this is an improvement.

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,401 @@ def __call__(self) -> Dict[str, str]:
...
-        credentials_provider=posit_strategy.sql_credentials_provider(cfg),
-    ) as connection:
-        with connection.cursor() as cursor:
-            cursor.execute(query)
-            rows = cursor.fetchall()
-            print(pd.DataFrame([row.asDict() for row in rows]))
+    @render.text
+    def text():
+        databricks_user_info = CurrentUserAPI(ApiClient(cfg())).me()
+        return f"Hello, {databricks_user_info.display_name}!"
```
tdstein, line 134: nit; extract variables to improve readability.

```suggestion
        current_user_api = CurrentUserAPI(ApiClient(cfg()))
        databricks_user_info = current_user_api.me()
        return f"Hello, {databricks_user_info.display_name}!"
```

src/posit/connect/external/databricks.py
```diff
@@ -59,76 +53,14 @@ def _new_bearer_authorization_header(credentials: Credentials) -> Dict[str, str]
...
-            },
-        )
-        response.raise_for_status()
-
-        credentials = Credentials(**response.json())
-        return _new_bearer_authorization_header(credentials)
-
-
-class PositContentCredentialsProvider:
+class _PositConnectContentCredentialsProvider:
```
tdstein, line 56: Should this inherit from `CredentialsProvider`?

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,401 @@ def __call__(self) -> Dict[str, str]:
...
     """
 
     def __init__(
         self,
-        local_strategy: CredentialsStrategy,
         client: Optional[Client] = None,
+        user_session_token: Optional[str] = None,
     ):
-        self._local_strategy = local_strategy
+        self._cp = None
```
tdstein, line 237: nit; add a type signature for clarity on the non-null state.

```suggestion
        self._cp: CredentialsProvider | None = None
```

src/posit/connect/external/databricks.py
```diff
@@ -8,38 +8,32 @@
...
-    @abc.abstractmethod
-    def auth_type(self) -> str:
-        raise NotImplementedError
+POSIT_OAUTH_INTEGRATION_AUTH_TYPE = "posit-oauth-integration"
+POSIT_WORKBENCH_AUTH_TYPE = "posit-workbench"
 
-    @abc.abstractmethod
-    def __call__(self, *args, **kwargs) -> CredentialsProvider:
-        raise NotImplementedError
+logger = logging.getLogger("posit.sdk")
```
tdstein, line 36: I'm pretty sure we can use `logger = logging.getLogger(__name__)` here. Which will be equivalent to `getLogger("posit.sdk.external.databricks")`.

https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

tdstein (commented):
Looking good! Just a few minor questions.

nealrichardson:
This still feels like a lot of complexity for an app developer to have to wrestle with (or, more likely, just cargo cult):

```
    return databricks_config(
        posit_default_strategy=databricks_cli,
        posit_workbench_strategy=WorkbenchStrategy(),
        posit_connect_strategy=ConnectStrategy(user_session_token=session_token),
        host=os.getenv("DATABRICKS_HOST"),
        warehouse_id=os.getenv("DATABRICKS_WAREHOUSE_ID"),
    )
```

What if instead it were

```
    return databricks_config(user_session_token = session_token)
```

and everything else were set by defaults in the function signature? Seems totally reasonable to use conventions for env vars for defaults, and as for `posit_default_strategy` and `posit_workbench_strategy`, would anyone ever want something other than the default you provided? (If so, they could still pass it as an argument.)

tdstein:
> This still feels like a lot of complexity for an app developer to have to wrestle with

Agreed. You have all the building blocks now. Applying sensible defaults to `databricks_config` would be a nice UX.

dbkegley:
> This still feels like a lot of complexity for an app developer to have to wrestle with (or, more likely, just cargo cult):
> 
> ```
>     return databricks_config(
>         posit_default_strategy=databricks_cli,
>         posit_workbench_strategy=WorkbenchStrategy(),
>         posit_connect_strategy=ConnectStrategy(user_session_token=session_token),
>         host=os.getenv("DATABRICKS_HOST"),
>         warehouse_id=os.getenv("DATABRICKS_WAREHOUSE_ID"),
>     )
> ```
> 
> What if instead it were
> 
> ```
>     return databricks_config(user_session_token = session_token)
> ```
> 
> and everything else were set by defaults in the function signature? Seems totally reasonable to use conventions for env vars for defaults, and as for `posit_default_strategy` and `posit_workbench_strategy`, would anyone ever want something other than the default you provided? (If so, they could still pass it as an argument.)

That's basically how this is implemented with a minor difference. The example in the PR description shows how to override the default values. The env vars also also already handled the way that you describe, but it's useful to show them in our docs because if they aren't set then you get a rather opaque error message from Databricks SDK/SQL.

The minimal viewer auth example would be:

```python
return databricks_config(
  posit_connect_strategy=ConnectStrategy(user_session_token=session_token)
)
```

And the minimal service account auth example would be:

```python
return databricks_config()
```

However note that the minimal examples would not work when Workbench-managed Databricks credentials are used. You have to enable it by passing `WorkbenchStrategy(). There's some awkwardness with how the Databricks Config object is initialized.

tdstein (approved):

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,407 @@ def __call__(self) -> Dict[str, str]:
...
+
+    @reactive.calc
+    def cfg():
+        session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
+        return databricks_config(
+            posit_default_strategy=databricks_cli,
+            posit_workbench_strategy=WorkbenchStrategy(),
+            posit_connect_strategy=ConnectStrategy(user_session_token=session_token),
+            host=os.getenv("DATABRICKS_HOST"),
+            warehouse_id=os.getenv("DATABRICKS_WAREHOUSE_ID"),
```
plascaray, line 375:
```suggestion
```

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,407 @@ def __call__(self) -> Dict[str, str]:
...
+    @reactive.calc
+    def cfg():
         session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
-        posit_strategy = PositCredentialsStrategy(
-            local_strategy=databricks_cli, user_session_token=session_token
+        return databricks_config(
+            posit_workbench_strategy=oauth_service_principal,
+            posit_connect_strategy=ConnectStrategy(user_session_token=session_token),
+            host=os.getenv("DATABRICKS_HOST"),
+            warehouse_id=os.getenv("DATABRICKS_WAREHOUSE_ID"),
```
plascaray, line 416:
```suggestion
```
I don't think this is needed for the example to work.

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,407 @@ def __call__(self) -> Dict[str, str]:
...
-        # If the content is not running on Connect then fall back to local_strategy
-        if is_local():
-            return self._local_strategy(*args, **kwargs)
+    @reactive.calc
+    def cfg():
+        return databricks_config(
+            posit_workbench_strategy=azure_service_principal,
+            posit_connect_strategy=ConnectStrategy(),
+            host=os.getenv("DATABRICKS_HOST"),
+            warehouse_id=os.getenv("DATABRICKS_WAREHOUSE_ID"),
```
plascaray, line 458:
```suggestion
```
I don't think this is needed for the example to work.

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,407 @@ def __call__(self) -> Dict[str, str]:
...
+    This examples uses environment variables to load the correct configurations at runtime. Make sure to set
+    the `DATABRICKS_CLIENT_ID` and `DATABRICKS_CLIENT_SECRET` as environment variables in Posit Workbench, but do
+    not set them inside of Posit Connect.
 
-    app_ui = ui.page_fluid(ui.output_text("text"), ui.output_data_frame("result"))
+    ```python
+    import os
+
+    from databricks.sdk.core import ApiClient
+    from databricks.sdk.core.credentials_provider import oauth_service_principal
```
plascaray, line 398:
```suggestion
    from databricks.sdk.core import ApiClient, oauth_service_principal
```

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,407 @@ def __call__(self) -> Dict[str, str]:
...
+    This example shows how to construct a databricks Config that is compatible with:
+    - Databricks CLI authentication for local development
+    - Workbench-managed Databricks Credentials in Posit Workbench
+    - Viewer OAuth integration authentication in Posit Connect
+
+    ```python
+    import os
+
+    from databricks.sdk.core import ApiClient
+    from databricks.sdk.core.credentials_provider import databricks_cli
```
plascaray, line 355:
```suggestion
    from databricks.sdk.core import ApiClient, databricks_cli
```

src/posit/connect/external/databricks.py
```diff
@@ -160,277 +91,407 @@ def __call__(self) -> Dict[str, str]:
...
-        """The sql connector attempts to call the credentials provider w/o any args.
+    ```python
+    import os
 
-        The SQL client's `ExternalAuthProvider` is not compatible w/ the SDK's implementation of
-        `CredentialsProvider`, so create a no-arg lambda that wraps the args defined by the real caller.
-        This way we can pass in a databricks `Config` object required by most of the SDK's `CredentialsProvider`
-        implementations from where `sql.connect` is called.
+    from databricks.sdk.core import ApiClient
+    from databricks.sdk.core.credentials_provider import azure_service_principal
```
plascaray, line 441:
```suggestion
    from databricks.sdk.core import ApiClient, azure_service_principal
```

dbkegley:
@tdstein FYI - I've moved the workbench credentials helper into a new module `posit.workbench.external`

Planning to merge this later today and update the release notes for the next release.

PR #386, opened 2025-02-18
test: use ghcr image repository

tdstein:
Updates CI to use GHCR image repository instead of Docker Hub.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,6 +65,11 @@ jobs:
           - 2022.11.0
     steps:
       - uses: actions/checkout@v4
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


toph-allen (commented):
Looks good! (CI just failed because of conventional commits -- commit message (or is it PR title) was `tests` and not `test`)

[edit] wait the pr message IS `test` so disregard the above, looks good

toph-allen (approved):

nealrichardson:
I thought you didn't need this because the test suite is only using the public images? That's why there was no step that logs into Docker Hub.

tdstein:
> I thought you didn't need this because the test suite is only using the public images? That's why there was no step that logs into Docker Hub.

Ah, I must have misunderstood your message. I thought all images were to be removed from Docker Hub.

nealrichardson:
Nah, we aren't dealing with the public images yet, only the internal, private dev images. The public ones are harder since they're, well, public, and URLs are forever. Though even if/when we were to move those somewhere else, they'd still be public, so you wouldn't need to log in to access them.

PR #390, opened 2025-02-24
feat: support request body forwarding on redirects

### LINKED ISSUE #385: feat: forward POST body with 3xx redirection statues
When accessing Connect insecurely (via HTTP), POSTs respond with 3xx error codes. By default, the original POST body is not forwarded on redirection. Other packages have handled this by overriding the default handler to forward the POST body on redirection.

See https://github.com/rstudio/rsconnect/blob/a3ebf70adb96a020fb8844b7194870ae04500a1d/R/http.R#L118-L120

tdstein:
All requests sent over HTTP to Connect are responded to with a redirect to the HTTPS location. By default, `Requests` follows the default HTTP specification and removes the POST body when redirecting. We want to override this behavior to ensure the POST body is carried through the redirect.

Resolves #385

### DIFF
--- a/src/posit/connect/permissions.py
+++ b/src/posit/connect/permissions.py
@@ -158,7 +158,6 @@ class Permissions(Resources):
             kwargs["principal_type"] = principal_type
 
         path = f"v1/content/{self.content_guid}/permissions"
-        print(kwargs)
         response = self._ctx.client.post(path, json=kwargs)
         return Permission(self._ctx, **response.json())
 
--- a/src/posit/connect/permissions.py
+++ b/src/posit/connect/permissions.py
@@ -158,6 +158,7 @@ class Permissions(Resources):
             kwargs["principal_type"] = principal_type
 
         path = f"v1/content/{self.content_guid}/permissions"
+        print(kwargs)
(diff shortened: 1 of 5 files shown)


src/posit/connect/permissions.py
```diff
@@ -158,6 +158,7 @@ def create(
             kwargs["principal_type"] = principal_type
 
         path = f"v1/content/{self.content_guid}/permissions"
+        print(kwargs)
```
schloerke, line 161:
```suggestion
```

tests/posit/connect/test_sessions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    assert response.status_code == 200
+    assert len(responses.calls) == 2
+    # The initial call is a POST, but the follow-up should be a GET since preserve_post is False
+    assert responses.calls[0].request.method == "POST"
+    assert responses.calls[1].request.method == "GET"
+
+
+@responses.activate
+def test_post_redirect_307():
```
dbkegley, line 58: Do we need this test for 308 as well?

dbkegley (approved):

PR #391, opened 2025-02-26
feat: Add Posit product environment check helpers

### LINKED ISSUE #389: RSTUDIO_PRODUCT is deprecated; support POSIT_PRODUCT
This calls for helper(s) to determine which environment is being used. For example:
- `client.is_running_on_connect()`
- `client.is_running_on_workbench()`

or something similar to encapsulate the check.

Need to check both `RSTUDIO_PRODUCT` and `POSIT_PRODUCT` since older versions of connect and workbench will not have the latter set, but when it is set, they will be the same value. RSTUDIO_PRODUCT has not yet been removed, only deprecated for now.

Once merged, ensure that any documentation examples in our products use these helpers for the latest version of this utility. Add fallback note if not on version that includes these helpers.

mconflitti-pbc:
Fixes #389

### DIFF
--- a/src/posit/environment.py
+++ b/src/posit/environment.py
@@ -1,0 +1,30 @@
+from __future__ import annotations
+
+import os
+
+
+def get_product() -> str | None:
+    """Returns the product name if called with a Posit product.
+
+    The products will always set the environment variable `POSIT_PRODUCT=<product-name>`
+    or `RSTUDIO_PRODUCT=<product-name>`.
+
+    RSTUDIO_PRODUCT is deprecated and acts as a fallback for backwards compatibility.
+    It is recommended to use POSIT_PRODUCT instead.
+    """
+    return os.getenv("POSIT_PRODUCT") or os.getenv("RSTUDIO_PRODUCT")
+
+
+def is_local() -> bool:
+    """Returns true if called while running locally."""
+    return get_product() is None
+
+
+def is_running_on_connect() -> bool:
+    """Returns true if called from a piece of content running on a Connect server."""
+    return get_product() == "CONNECT"
+
+
+def is_running_on_workbench() -> bool:
+    """Returns true if called from within a Workbench server."""
+    return get_product() == "WORKBENCH"
--- a/src/posit/connect/_utils.py
+++ b/src/posit/connect/_utils.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
-import os
+import warnings
 
 from typing_extensions import Any
+
+from ..environment import is_local as env_is_local
 
 
 def update_dict_values(obj: dict[str, Any], /, **kwargs: Any) -> None:
@@ -33,10 +35,15 @@ def update_dict_values(obj: dict[str, Any], /, **kwargs: Any) -> None:
 
 
 def is_local() -> bool:
-    """Returns true if called from a piece of content running on a Connect server.
-
-    The connect server will always set the environment variable `RSTUDIO_PRODUCT=CONNECT`.
-    We can use this environment variable to determine if the content is running locally
-    or on a Connect server.
     """
-    return os.getenv("RSTUDIO_PRODUCT") != "CONNECT"
+    Check if code is running locally.
+
+    .. deprecated:: 0.9.0
+        Use :func:`posit.environment.is_local` instead.
+    """
+    warnings.warn(
+        "posit.connect._utils.is_local is deprecated. Use posit.environment.is_local instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return env_is_local()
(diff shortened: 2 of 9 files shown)


src/posit/environment.py
tdstein, line 1: There's also https://github.com/posit-dev/posit-sdk-py/blob/main/src/posit/connect/environments.py. Would `runtime.py` be a better fit for this functionality?

src/posit/environment.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def is_running_on_connect() -> bool:
+    """Returns true if called from a piece of content running on a Connect server."""
+    return get_product() == "CONNECT"
+
+
+def is_running_on_workbench() -> bool:
+    """Returns true if called from within a Workbench server."""
+    return get_product() == "WORKBENCH"
```
tdstein, line 30: Do we need to support dynamic environment variables here? Another option is to use static variables.

```python

import os

POSIT_PRODUCT = os.getenv("POSIT_PRODUCT") or os.getenv("RSTUDIO_PRODUCT")
IS_POSIT_CONNECT = POSIT_PRODUCT == "CONNECT"
IS_POSIT_WORKBENCH = POSIT_PRODUCT == "WORKBENCH"
```

Then from `external.databricks/snowflake`...

```python
from .. import runtime

if not runtime.POSIT_PRODUCT:
    return;
```

Feel free to disregard this if you have already thought through it.

tdstein (commented):
Everything looks good. Thanks for adding the deprecation warning! Ready to approve, but want to get your thoughts on `runtime` vs `environment` for the module name.

src/posit/environment.py
```diff
@@ -0,0 +1,30 @@
...
+    The products will always set the environment variable `POSIT_PRODUCT=<product-name>`
+    or `RSTUDIO_PRODUCT=<product-name>`.
+
+    RSTUDIO_PRODUCT is deprecated and acts as a fallback for backwards compatibility.
+    It is recommended to use POSIT_PRODUCT instead.
+    """
+    return os.getenv("POSIT_PRODUCT") or os.getenv("RSTUDIO_PRODUCT")
+
+
+def is_local() -> bool:
```
tdstein, line 18: I suppose `is_local` is a little subjective since their application could run elsewhere that isn't their machine or Posit.

src/posit/environment.py
mconflitti-pbc: great suggestion!

src/posit/environment.py
mconflitti-pbc: this is great feedback. did not consider this.

i think i could go either way here. your approach is definitely simpler. let me ensure its testable. i would think just importing the module should trigger this to run but can check.

The connectapi and connectcreds equivalents are functions so followed suit but i like simple :)

mconflitti-pbc:
@zackverham and @dbkegley after looking at the existing is_local function, it was only checking if POSIT_PRODUCT is not set to CONNECT. So it would be considered local if running on workbench. Is that the intended effect?

src/posit/environment.py
dbkegley: I think this might be a little misleading for users today since workbench doesn't support this var yet.

I haven't tested this on workbench yet but I'm expecting to do something like this:

```
https://github.com/posit-dev/posit-sdk-py/pull/384/files#diff-08b18171119ae963aeaa57073337c0ad9df56da3f6c90cd887c3245bd7b556feR35-R45
```

mconflitti-pbc: I think the functional approach is going to best here, especially from a testing perspective. Complicates things when the values are set upon import. Going to keep it functional.

@dbkegley yeah that is fine! can add these checks into the workbench runtime check

mconflitti-pbc: @atheriel, do you have any thoughts here?

mconflitti-pbc:
> > @zackverham and @dbkegley after looking at the existing is_local function, it was only checking if POSIT_PRODUCT is not set to CONNECT. So it would be considered local if running on workbench. Is that the intended effect?
> 
> That was the intent initially but now that we need a better way to differentiate between Connect vs Workbench vs elsewhere the original implementation no longer seems correct.

Okay cool. This PR should address that then to ensure it is not connect nor workbench to be local.

src/posit/environment.py
mconflitti-pbc: going to keep this since staying with functional approach

src/posit/environment.py
mconflitti-pbc: Always the option punt and not have any workbench related check for now.

tdstein:
@mconflitti-pbc - are we still trying to move forward with this?

tdstein:
Should we close this PR and open a new one when ready? Changing it back to a draft also works.

PR #394, opened 2025-02-28
chore: Add `*.lic` to `.gitignore`

jonkeane:
Adding `*.lic` to `.gitignore` so that accidentally committing license files requires additional steps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,3 +170,6 @@ cython_debug/
 
 /.luarc.json
 _dev/
+
+# license files should not be commited to this repository
+*.lic
--- a/integration/.gitignore
+++ b/integration/.gitignore
@@ -1,3 +1,2 @@
-*.lic
 logs
 reports


.gitignore
```diff
@@ -170,3 +170,6 @@ cython_debug/
 
 /.luarc.json
 _dev/
+
+# license files should not be commited to this repository
+*.lic
```
tdstein, line 175: This also exists here: https://github.com/posit-dev/posit-sdk-py/blob/56c6eed21a7d2ca3abb9d9bccb5986f097cb77a2/integration/.gitignore#L1

The integration tests expect it to exist in the integration directory.

Screenshot 2025-02-28 at 15 36 14

tdstein (approved):

.gitignore
jonkeane: Ah, good catch. Would you rather only have it at the top level (since that syntax will catch all instances no matter where they are) or in both places? Either is fine, both places is a little bit silly, but makes the intent clear(er)?

tdstein: The top level is better since it shouldn't be checked in anywhere.

jonkeane: Done, thanks!

PR #396, opened 2025-03-11
feat: speed up the packages api

tdstein:
The Connect Packages API (v1/packages) supports uncapped page size. By increasing the page size from the default of 500 to 1,000,000, there is a significant speed up when fetching all packages. This is due to a reduction in the number of round trips to the Connect server. The tradeoff is a larger payload. Since the Python requests package does not limit the  HTTP GET request-response body size, we are unconcerned with the larger payload.


toph-allen (approved):
Looks good to me!

PR #397, opened 2025-03-11
feat: added standard token type enum; added aws_creds helper

mconflitti-pbc:
- Refactored supported token types into a string enum
- Added aws creds helper to properly decode the value from the Connect credential exchange endpoint

### DIFF
--- a/tests/posit/connect/oauth/test_oauth.py
+++ b/tests/posit/connect/oauth/test_oauth.py
@@ -4,7 +4,7 @@ import pytest
 import responses
 
 from posit.connect import Client
-from posit.connect.oauth.oauth import API_KEY_TOKEN_TYPE, _get_content_session_token
+from posit.connect.oauth.oauth import OAuthTokenType, _get_content_session_token
 
 
 class TestOAuthIntegrations:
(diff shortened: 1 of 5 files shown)


tests/posit/connect/oauth/test_oauth.py
```diff
@@ -62,11 +62,38 @@ def test_get_credentials_api_key(self):
...
         c = Client(api_key="12345", url="https://connect.example/")
         c._ctx.version = None
-        creds = c.oauth.get_credentials("cit", API_KEY_TOKEN_TYPE)
+        creds = c.oauth.get_credentials("cit", OAuthTokenType.API_KEY)
         assert creds.get("access_token") == "viewer-api-key"
         assert creds.get("issued_token_type") == "urn:posit:connect:api-key"
         assert creds.get("token_type") == "Key"
 
+    @responses.activate
+    def test_get_credentials_aws(self):
```
zackverham, line 71: maybe I'm missing something - does this duplicate the tests in `test_aws.py`?

zackverham (approved):
This approach looks reasonable to me!

PR #399, opened 2025-03-21
feat: Add support for AWS integration content session token exchange

mconflitti-pbc:
Similar to existing helper for Viewer integrations but the content session token counterpart. 

Usage example in docstring for function.

Tests updated.

### DIFF
--- a/src/posit/connect/external/aws.py
+++ b/src/posit/connect/external/aws.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import base64
 import json
+from datetime import datetime
 
-from typing_extensions import TYPE_CHECKING, Dict, Optional
+from typing_extensions import TYPE_CHECKING, Optional, TypedDict
 
 from ..oauth.oauth import OAuthTokenType
 
@@ -11,13 +12,20 @@ if TYPE_CHECKING:
     from ..client import Client
 
 
-def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, str]:
+class Credentials(TypedDict):
+    aws_access_key_id: str
+    aws_secret_access_key: str
+    aws_session_token: str
+    expiration: datetime
+
+
+def get_credentials(client: Client, user_session_token: str) -> Credentials:
     """
     Get AWS credentials using OAuth token exchange for an AWS Viewer integration.
 
-    According to RFC 8693, the access token must be a base64 encoded JSON object
-    containing the AWS credentials. This function will decode and deserialize the
-    access token and return the AWS credentials.
+    According to RFC 8693, the access token must be a base64-encoded JSON object
+    containing the AWS credentials. This function will return the decoded and
+    deserialized AWS credentials.
 
     Examples
     --------
@@ -32,9 +40,9 @@ def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, st
     credentials = get_aws_credentials(client, user_session_token)
     aws_session_expiration = credentials["expiration"]
     aws_session = boto3.Session(
-        aws_access_key_id=credentials["accessKeyId"],
-        aws_secret_access_key=credentials["secretAccessKey"],
-        aws_session_token=credentials["sessionToken"],
+        aws_access_key_id=credentials["aws_access_key_id"],
+        aws_secret_access_key=credentials["aws_secret_access_key"],
+        aws_session_token=credentials["aws_session_token"],
     )
 
     s3 = aws_session.resource("s3")
@@ -67,15 +75,15 @@ def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, st
     return _decode_access_token(access_token)
 
 
-def get_aws_content_credentials(
+def get_content_credentials(
     client: Client, content_session_token: Optional[str] = None
-) -> Dict[str, str]:
+) -> Credentials:
     """
     Get AWS credentials using OAuth token exchange for an AWS Service Account integration.
 
-    According to RFC 8693, the access token must be a base64 encoded JSON object
-    containing the AWS credentials. This function will decode and deserialize the
-    access token and return the AWS credentials.
+    According to RFC 8693, the access token must be a base64-encoded JSON object
+    containing the AWS credentials. This function will return the decoded and
+    deserialized AWS credentials.
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


src/posit/connect/external/aws.py
```diff
@@ -13,12 +13,34 @@
...
+    ```python
+    from posit.connect import Client
+    from posit.connect.external.aws import get_aws_credentials
+    import boto3
+    from shiny.express import session
+
+    client = Client()
+    session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
+    credentials = get_aws_credentials(client, user_session_token)
+    aws_session_expiration = credentials["expiration"]
```
zackverham, line 33: nit: I don't think this variable actually gets used anywhere

src/posit/connect/external/aws.py
```diff
@@ -42,6 +64,84 @@ def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, st
...
+    Examples
+    --------
+    ```python
+    from posit.connect import Client
+    from posit.connect.external.aws import get_aws_content_credentials
+    import boto3
+
+    client = Client()
+    credentials = get_aws_content_credentials(client)
+    session_expiration = credentials["expiration"]
```
zackverham, line 89: nit: I don't think this variable gets used anywhere

src/posit/connect/external/aws.py
```diff
@@ -42,6 +64,84 @@ def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, st
...
+
+    Examples
+    --------
+    ```python
+    from posit.connect import Client
+    from posit.connect.external.aws import get_aws_content_credentials
+    import boto3
+
+    client = Client()
+    credentials = get_aws_content_credentials(client)
```
zackverham, line 88: should we show getting the content session token from an env var here? Or do we expect the client's ability to handle that behind the scenes to imply that customer's don't need to see that bit?

src/posit/connect/external/aws.py
```diff
@@ -13,12 +13,34 @@
 
 def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, str]:
     """
-    Get AWS credentials using OAuth token exchange.
+    Get AWS credentials using OAuth token exchange for an AWS Viewer integration.
 
     According to RFC 8693, the access token must be a base64 encoded JSON object
```
zackverham, line 18: nit in these docstrings: I think base64 encoded should be hyphenated (base64-encoded). 

Obviously not remotely a big deal, just saw it as I was reading

zackverham (approved):
Couple of tiny observations / nits, but the overall pattern matches what we've done previously so I think this looks great. 

Probably will want Taylor to give it a lookover as well just to make sure he doesn't see anything I missed.

src/posit/connect/external/aws.py
```diff
@@ -13,12 +13,34 @@
...
+
+    client = Client()
+    session_token = session.http_conn.headers.get("Posit-Connect-User-Session-Token")
+    credentials = get_aws_credentials(client, user_session_token)
+    aws_session_expiration = credentials["expiration"]
+    aws_session = boto3.Session(
+        aws_access_key_id=credentials["accessKeyId"],
+        aws_secret_access_key=credentials["secretAccessKey"],
+        aws_session_token=credentials["sessionToken"],
+    )
```
tdstein, line 38: We should return a `TypedDict` here and map the keys to the Session keyword arguments:

```python
class Credentials(TypedDict):
    aws_access_key_id: str
    aws_secret_access_key: str
    aws_session_token: str
```

This saves the users a few steps...

```suggestion
    credentials = get_aws_credentials(client, user_session_token)
    aws_session = boto3.Session(**credentials)
```

src/posit/connect/external/aws.py
```diff
@@ -42,6 +64,84 @@ def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, st
     access_token = credentials.get("access_token")
     if not access_token:
         raise ValueError("No access token found in credentials")
+    return _decode_access_token(access_token)
+
+
+def get_aws_content_credentials(
```
tdstein, line 70: When do I use `get_aws_content_credentials` instead of `get_aws_credentials`?

src/posit/connect/external/aws.py
mconflitti-pbc: ah right. just wanted to show that this exists in the returned object

src/posit/connect/external/aws.py
mconflitti-pbc: I think that this is sufficient, given that this already needs to be guarded by checking if this is running in connect. That may be a good thing to add in the example: on_connect()

src/posit/connect/external/aws.py
mconflitti-pbc: content credentials are used for service account integrations. the other is for viewer integrations.

src/posit/connect/external/aws.py
```diff
@@ -13,12 +13,34 @@
 
 def get_aws_credentials(client: Client, user_session_token: str) -> Dict[str, str]:
```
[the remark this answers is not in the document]
mconflitti-pbc: ah good point

src/posit/connect/external/aws.py
mconflitti-pbc: will change this slightly since this will be changed to a typeddict

src/posit/connect/external/aws.py
tdstein: Should we add that to the docstring here? Or is well enough known when reading the user guide? I'm probably missing some context.

src/posit/connect/external/aws.py
mconflitti-pbc: probably can't just pass it through though since the credentials includes the expiration field which the sesison constructor does not know about and will error out

tdstein: Ah, that makes sense. Probably best not to tightly couple it anyway in case boto3 changes down the road. What are your thoughts on using the TypedDict with snake_case variable names and then keeping the rest the same?

mconflitti-pbc: Yeah true good point. Yes will do that

tdstein (approved):

PR #404, opened 2025-05-01
feat: add method to retrieve active bundle

### LINKED ISSUE #380: Add method to retrieve "active" bundle.
Add a helper method to the `Bundles` class, retrieving the "active" bundle. This is determined by checking the `active` field on the bundle object. Currently, the this can only be done by retrieving the full bundle list and then filtering for the match.

```python
client = connect.Client()
content = client.content.get(guid)
bundle = content.bundles.active()
```

tdstein:
Adds `active()` method to `Bundles` class to retrieve the currently active bundle.

Resolves #380

### DIFF
--- a/.github/workflows/site.yaml
+++ b/.github/workflows/site.yaml
@@ -24,8 +24,8 @@ jobs:
           fetch-depth: 0
       - uses: astral-sh/setup-uv@v3
       - run: uv python install
-      - uses: quarto-dev/quarto-actions/setup@v2
       - run: make build install
+      - uses: quarto-dev/quarto-actions/setup@v2
       - run: make docs
       - uses: actions/configure-pages@v3
       - uses: actions/upload-pages-artifact@v3
--- a/.github/workflows/site.yaml
+++ b/.github/workflows/site.yaml
@@ -24,8 +24,8 @@ jobs:
           fetch-depth: 0
       - uses: astral-sh/setup-uv@v3
       - run: uv python install
-      - run: make build install
       - uses: quarto-dev/quarto-actions/setup@v2
+      - run: make build install
       - run: make docs
       - uses: actions/configure-pages@v3
       - uses: actions/upload-pages-artifact@v3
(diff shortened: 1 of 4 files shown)


tdstein:
I did a side quest using Claude Code.

.github/workflows/site.yaml
```diff
@@ -24,8 +24,8 @@ jobs:
           fetch-depth: 0
       - uses: astral-sh/setup-uv@v3
       - run: uv python install
-      - run: make build install
       - uses: quarto-dev/quarto-actions/setup@v2
+      - run: make build install
```
jonkeane, line 28: Is this change intensional?

integration/tests/posit/connect/test_bundles.py
```diff
@@ -0,0 +1,99 @@
...
+        assert bundle["content_guid"] == self.content["guid"]
+
+    def test_get_bundle(self):
+        """Test getting a specific bundle."""
+        # Create a bundle first
+        created_bundle = self.content.bundles.create(str(self.bundle_path))
+
+        # Get the bundle by ID
+        bundle = self.content.bundles.get(created_bundle["id"])
+        assert bundle["id"] == created_bundle["id"]
```
jonkeane, line 63: These were all tests for functionality we had already but was not (yet) tested, yeah?

tdstein: Correct.

.github/workflows/site.yaml
tdstein: Yeah, trying to figure out what the GH Action is broken...

tdstein: See https://github.com/posit-dev/posit-sdk-py/pull/407

jonkeane: 🙏 thanks!

jonkeane (approved):
Thanks for this! I do appreciate that the operative code is small and we're adding test cases for places we lacked some before.

PR #405, opened 2025-05-01
feat: update responses with content_guid and content_id

tdstein:


tdstein:
Closing as no longer relevant — the SDK should match the REST API as-is.

PR #408, opened 2025-05-06
feat: improve task polling with exponential backoff

tdstein:
Implements an improved task polling mechanism with configurable exponential backoff to reduce the number of API calls for long-running tasks. This helps to minimize resource consumption on the Connect server while still providing responsive feedback.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/posit/connect/tasks.py
+++ b/src/posit/connect/tasks.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import time
+
 from typing_extensions import overload
 
 from . import resources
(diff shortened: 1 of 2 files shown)


src/posit/connect/tasks.py
```diff
@@ -95,17 +97,40 @@ def update(self, *args, **kwargs) -> None:
...
+        while waiting for long-running tasks. The first request uses the initial_wait value,
+        and subsequent requests increase the wait time by the backoff factor, up to max_wait. To disable exponential backoff, set backoff to 1.0.
         """
+        wait_time = initial_wait
+
         while not self.is_finished:
             self.update()
 
+            # Wait client-side
+            time.sleep(wait_time)
```
aronatkins, line 129: I was trying to suggest using the `wait=N` query argument to implement long-polling and not do any sleeping on the client.

mconflitti-pbc: Given that the sdk's http client is not async, I am curious what relying on long polling would gain here. Could be mistaken but whether you have a long standing connection or sleep, both are blocking until the response is returned right? If that is accurate then having the blocking driven by the client may still give this a slight edge.

aronatkins: when you call `/tasks/DECAFBAD?wait=10`, the API call will wait up to ten seconds or until the task is complete before returning. read that call as "accumulate data for up to 10 seconds before returning the result".

the call returns as soon as the task is complete.

when the client is sleeping, it has to wait the entire duration before it can make an API call to fetch additional data. if the task completes after 1s, you still need to wait the remaining 9s.

tdstein: Ah, I completely spaced on the API functionality for sleeps. I'll modify the implementation to use it.

tdstein: Much better :) https://github.com/posit-dev/posit-sdk-py/pull/409

PR #409, opened 2025-05-06
feat: improve task polling

### LINKED ISSUE #184: Improve task polling
The existing task polling is naive and doesn't consider the various edge cases around output length, polling algorithms, etc...

Modify the implementation to take advantage of the log appending support in the API. 
Modify the polling algorithm to reduce server resource consumption (i.e., do polling client side instead of server side). 
Check if there is a need for supporting more advanced polling algorithms (.e.g, backoff policies, etc).

tdstein:
Implements an improved task polling mechanism with configurable wait time. This helps to minimize resource consumption on the Connect server while still providing responsive feedback.

Resolves #184

### DIFF
--- a/src/posit/connect/tasks.py
+++ b/src/posit/connect/tasks.py
@@ -97,39 +97,21 @@ class Task(resources.BaseResource):
         result = response.json()
         super().update(**result)
 
-    def wait_for(self, *, initial_wait: int = 1, max_wait: int = 10, backoff: float = 1.5) -> None:
+    def wait_for(self, *, wait: int = 1) -> None:
         """Wait for the task to finish.
 
         Parameters
         ----------
-        initial_wait : int, default 1
-            Initial wait time in seconds. First API request will use this as the wait parameter.
-        max_wait : int, default 10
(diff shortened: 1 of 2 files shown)


src/posit/connect/tasks.py
```diff
@@ -97,39 +95,21 @@ def update(self, *args, **kwargs) -> None:
...
         Parameters
         ----------
-        initial_wait : int, default 1
-            Initial wait time in seconds. First API request will use this as the wait parameter.
-        max_wait : int, default 10
-            Maximum wait time in seconds between polling requests.
-        backoff : float, default 1.5
-            Backoff multiplier for increasing wait times.
+        wait : int, default 1
+            Wait time in seconds between polling requests.
```
aronatkins, line 104:
```suggestion
            Maximum amount of time for the task API request to capture status changes.
```

aronatkins: I don't love my suggested wording, either.

aronatkins (approved):

tdstein:
Added a maximum attempts option so folks can break out of the loop if desired.

PR #416, opened 2025-06-24
Extend content.get() to support `include` param

cgraham-rs:
Adds `include` param support for `content.get()`

- Add the param support
- Add unit test
- Add integration test
- Also manually verified by testing locally against a live Connect instance

### DIFF
--- a/integration/tests/posit/connect/test_content.py
+++ b/integration/tests/posit/connect/test_content.py
@@ -23,15 +23,17 @@ class TestContent:
         assert self.client.content.count() == 1
 
     def test_get(self):
-        assert self.client.content.get(self.content["guid"]) == self.content
-
-    def test_get_with_include_string(self):
-        assert self.client.content.get(self.content["guid"], include="owner")
-
-    def test_get_with_include_list(self):
-        assert self.client.content.get(
-            self.content["guid"], include=["owner", "tags", "vanity_url"]
-        )
+        item = self.client.content.get(self.content["guid"])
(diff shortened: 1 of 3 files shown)


tdstein:
I wonder if the better UX is to simply do all of the includes all the time in `get`. I can't think of any use cases where making this configurable is strictly necessary.

cgraham-rs:
> I wonder if the better UX is to simply do all of the includes all the time in `get`. I can't think of any use cases where making this configurable is strictly necessary.

That sounds reasonable to me and fits my specific use case perfectly.  I simply followed the Connect API where they're all optional.

cgraham-rs:
> Go for it. It should be just modifying the request to include the query arguments.

@tdstein Done. 

Note that CI cannot run integration tests from a fork due to security around secrets.

`connect-1  | time="2025-06-25T21:29:05.292Z" level=warning msg="Unable to obtain a valid license: Your Posit Connect license has expired. Please contact your Customer Success representative or email sales@posit.co to obtain a current license."`


Locally I ran `2025.03.0` and `2025.02.0` successfully, but `preview` fails a variety of other tests.

tdstein:
> Note that CI cannot run integration tests from a fork due to security around secrets.

@cgraham-rs - since you are using a fork, you don't have access to the repository secrets. If you set the `CONNECT_LICENSE` environment variable in GitHub secrets, it should work on your branch. 

Alternatively, you can recreate the pull request as a branch off of main. Either should work.

> preview fails a variety of other tests.

I'm also seeing that. I'll address it in as a separate issue.

PR #419, opened 2025-07-03
feat: add optional audience parameter to credential exchange related methods

mconflitti-pbc:
- Adds in support for the optional audience parameter for the credential exchange endpoint to determine which integration to use
- Removes "only 1 association" constraint on updating content's oauth integration associations.
- Adds `find_by` helper method for integrations and associations to allow a user to search available integrations (globally or scoped to a specific piece of content) to get the GUID for the integration they need dynamically.

### DIFF
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -2,9 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Optional
-
-from typing_extensions import TYPE_CHECKING, overload
+from typing_extensions import TYPE_CHECKING, Optional, overload
 
 from . import hooks, me
 from .auth import Auth
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -11,7 +11,8 @@ from .content import Content
 from .context import Context, ContextManager, requires
 from .groups import Groups
 from .metrics.metrics import Metrics
-from .oauth.oauth import OAuth, OAuthTokenType
+from .oauth.oauth import OAuth
+from .oauth.types import OAuthTokenType
 from .resources import _PaginatedResourceSequence, _ResourceSequence
 from .sessions import Session
 from .system import System
--- a/src/posit/connect/client.py
+++ b/src/posit/connect/client.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import Optional
+
 from typing_extensions import TYPE_CHECKING, overload
 
 from . import hooks, me
@@ -176,7 +178,11 @@ class Client(ContextManager):
         self._ctx = Context(self)
 
     @requires("2025.01.0")
-    def with_user_session_token(self, token: str) -> Client:
+    def with_user_session_token(
+        self,
+        token: str,
+        audience: Optional[str] = None,
+    ) -> Client:
         """Create a new Client scoped to the user specified in the user session token.
 
         Create a new Client instance from a user session token exchange for an api key scoped to the
@@ -256,7 +262,9 @@ class Client(ContextManager):
             raise ValueError("token must be set to non-empty string.")
 
         visitor_credentials = self.oauth.get_credentials(
-            token, requested_token_type=OAuthTokenType.API_KEY
+            token,
+            requested_token_type=OAuthTokenType.API_KEY,
+            audience=audience,
         )
 
         visitor_api_key = visitor_credentials.get("access_token", "")
--- a/src/posit/connect/content.py
+++ b/src/posit/connect/content.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import posixpath
 import time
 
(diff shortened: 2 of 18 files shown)


src/posit/connect/oauth/associations.py
```diff
@@ -66,9 +70,11 @@ def delete(self) -> None:
         path = f"v1/content/{self.content_guid}/oauth/integrations/associations"
         self._ctx.client.put(path, json=data)
 
-    def update(self, integration_guid: str) -> None:
+    def update(self, integration_guid: str | list[str]) -> None:
```
tdstein, line 73: In these situations, I've found that its better to do one or the other (str or list[str]).

src/posit/connect/oauth/oauth.py
```diff
@@ -89,10 +117,80 @@ def get_content_credentials(
...
+                if isinstance(name, re.Pattern):
+                    if not name.search(integration_name):
+                        match = False
+                else:
+                    if integration_name != name:
+                        match = False
+
+            if guid is not None and association.get("oauth_integration_guid") != guid:
+                match = False
+
```
tdstein, line 176: Is it possible to move this logic to `ContentItemAssociations(self._ctx, content_guid=content_guid).find_one()`?

`find_one` could call `find` internally and perform the client side filtering if the API doesn't support query parameters.

mconflitti-pbc: Due to the fact that we want to be able to support globally accessible integrations eventually, I moved this to a find_by function hanging off of client.oauth.integrations. The credentials endpoint will currently raise an error if that integration is not associated with the content but that will slowly change over time.

src/posit/connect/client.py
```diff
@@ -176,7 +177,11 @@ def __init__(self, *args, **kwargs) -> None:
         self._ctx = Context(self)
 
     @requires("2025.01.0")
-    def with_user_session_token(self, token: str) -> Client:
+    def with_user_session_token(
+        self,
+        token: str,
+        audience: Optional[str] = None,
+    ) -> Client:
```
dbkegley, line 184: Will there ever be >1 Connect viewer integration associated? If not then it might be nice to fetch the integration's guid internally so that the user doesn't need to specify the audience before initializing a viewer client.

mconflitti-pbc: I agree! So basically like, get the contents associations and grab the first visitor api key that shows up.

This works well in the cases where we would be requiring the visitor api key to associated with the content (which we could check for first automatically if no audience is provided), but we also want to be able to just access that particular integration in some general sense like: "connect-visitor-api-key" and we do that mapping in the backend. Then no association is needed in that special case which is desirable for extensions!

Curious your thoughts on that.

mconflitti-pbc: That's not a bad idea! I actually nixed the function using the `/v1/content/<guid>/associations` though. I added a find_by to integrations generally that can be used to pick the audience they want. This way we don't explicitly only check associations in the case someone wanted to use a globally accessible integration like that. Future proofs it slightly or at least doesnt make any assumptions quite yet as to where the guids come from.

mconflitti-pbc: I will follow up on this change once I have the backend changes in that would allow connect template integrations to be globally accessible. Then I can try to resolve the default connect api key integration that is created upon start

src/posit/connect/oauth/associations.py
```diff
@@ -59,16 +67,78 @@ def find(self) -> List[Association]:
             for result in response.json()
         ]
 
+    # TODO turn this on before merging
+    # @requires("2025.07.0")
```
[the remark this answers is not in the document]
mconflitti-pbc: I am forgetting now, does that automatically include higher versions than that as well or would that need a small PR to bump afterwards?

src/posit/connect/content.py
```diff
@@ -1000,3 +1001,16 @@ def get(self, guid: str) -> ContentItem:
 
         response = self._ctx.client.get(f"v1/content/{guid}", params=params)
         return ContentItem(self._ctx, **response.json())
+
+    @property
+    def current(self) -> ContentItem:
```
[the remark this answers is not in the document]
tdstein: Yeah, it's a special case. The default argument would be the environment variable.

```python
def get(self, uid = os.getenv("CONNECT_CONTENT_GUID"):
    if uid is None:
        raise ValueError("uid not set")
    ...
```

tdstein (approved):
Aside from the question on `associations.update`, this looks good to me.

src/posit/connect/oauth/associations.py
```diff
@@ -59,16 +67,71 @@ def find(self) -> List[Association]:
...
+
     def delete(self) -> None:
         """Delete integration associations."""
         data = []
 
         path = f"v1/content/{self.content_guid}/oauth/integrations/associations"
         self._ctx.client.put(path, json=data)
 
-    def update(self, integration_guid: str) -> None:
+    def update(self, integration_guids: list[str]) -> None:
```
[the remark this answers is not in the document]
tdstein: Oh right. That sounds like me 🙃 

Release notes seem ok.

PR #423, opened 2025-08-11
feat: auto discover audience for connect api integrations

mconflitti-pbc:
checks the current content's assoications and if there is a single connect api integration when there are multiple associations it will grab that guid automatically.


tdstein (approved):
Nice! You may try updating quartodoc to fix the CI failure.

PR #432, opened 2025-11-12
ci: adds Claude bot

tdstein:
Initial change to add Claude to this project.

### DIFF
--- a/.github/workflows/claude.yaml
+++ b/.github/workflows/claude.yaml
@@ -28,7 +28,7 @@ jobs:
       (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
       github.event_name == 'pull_request'
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 5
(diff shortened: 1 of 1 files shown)


.github/workflows/claude.yaml
```diff
@@ -0,0 +1,66 @@
...
+jobs:
+  default:
+    if: |
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
+      (github.event_name == 'issues' && contains(github.event.issue.body, '@claude')) ||
+      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
+      github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
nealrichardson: Watching this go on Connect and occasionally getting stuck, I think this timeout could be lower--if it succeeds, it only takes a few minutes.

PR #437, opened 2025-12-05
feat: add support for content session tokens in snowflake authenticator

mconflitti-pbc:
This is now relevant and needed given Snowflake adding support for oauth federation through an external idp: https://community.snowflake.com/s/article/Create-External-OAuth-Token-Using-Azure-AD-For-The-OAuth-Client-Itself

Simple change to support both session token types.


tdstein:
Nice! Did you get to test this in a Snowflake environment?

mconflitti-pbc:
@tdstein Will DM you the Connect docs PR. But yes!

PR #442, opened 2026-01-06
feat: adds APIs for Custom OAuth Integrations and add support getDelegatedAzureToken

zachhannum:
Addresses rstudio/rstudio-pro#9252 and rstudio/rstudio-pro#9679

This PR introduces a new Posit Workbench client for the Python SDK with comprehensive OAuth integration support.

## Example Usage

```python
from posit.workbench import Client

# Initialize client (uses Workbench session environment)
client = Client()

# List all OAuth integrations
integrations = client.oauth.integrations.find()

# Find a specific integration
github = client.oauth.integrations.find_by(name="^github-main$")

# Get credentials for an integration
creds = client.oauth.get_credentials(github["guid"])
access_token = creds["access_token"]

# Get Azure delegated token
azure_token = client.oauth.get_delegated_azure_token(
    "https://management.azure.com/"
)
```

### DIFF
--- a/src/posit/workbench/context.py
+++ b/src/posit/workbench/context.py
@@ -5,10 +5,9 @@ import functools
 # version stuff from RSTUDIO_VERSION
 import os
 import weakref
-from typing import Callable
 
 from packaging.version import Version
-from typing_extensions import TYPE_CHECKING, Concatenate, ParamSpec, Protocol, TypeVar
+from typing_extensions import TYPE_CHECKING, Callable, Concatenate, ParamSpec, Protocol, TypeVar
 
 if TYPE_CHECKING:
     from .client import Client


src/posit/workbench/client.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if not server_url:
+            raise EnvironmentError(
+                "RS_SERVER_ADDRESS environment variable not set. Cannot determine Posit Workbench server address. Contact your server administrator to ensure that launcher-sessions-callback-address is configured correctly.",
+            )
+        self.server_url: str = server_url
+        self._ctx = Context(self)
+
+    def get(self, path: str, **kwargs) -> Response:
+        url = self.server_url + path
```
tdstein: You may want to consider using `urllib` to make this URL building more robust. On the Connect side, we used https://github.com/posit-dev/posit-sdk-py/blob/workbench/zh-oauth-integrations/src/posit/connect/urls.py to handle some of the oddities around Connect URL expectations. Although, since your server_url is coming from RS_SERVER_ADDRESS that may not be needed. It's probably already well formed.

tdstein (commented):
Overall, I think this looks great! I'm super excited to see this addition to the project.

I think ideally it would good to have more unit tests around `cookie_reader.py` to help with long term stability. But, I'm not opposed to moving merging as is.

Additionally, have you considered adding integration tests? On the Connect side, I have found that extremely helpful for testing between server versions.

I'll give others a chance to review before approving. But let me know if there is urgency around getting this merged. The release process is mostly automated and should not take much time to prepare.

PR #459, opened 2026-03-12
feat: add has_content_credentials and return None from get_content_credentials

tdstein:
## Summary

- Adds `OAuth.has_content_credentials()` method so applications can check whether OAuth content credentials are available without manually inspecting `CONNECT_CONTENT_SESSION_TOKEN*` environment variables
- Changes `OAuth.get_content_credentials()` to return `None` instead of raising `ValueError` when no content session token is available
- Logs a warning when the token file exists but is empty

### Example usage

```python
from posit import connect

client = connect.Client()

if client.oauth.has_content_credentials():
    credentials = client.oauth.get_content_credentials()
    token = credentials["access_token"]

# Or simply:
credentials = client.oauth.get_content_credentials()
if credentials is not None:
    token = credentials["access_token"]
```

Closes #458

## Test plan

- [x] Existing tests updated for new return type (`None` instead of `ValueError`)
- [x] New tests for `has_content_credentials` (env var present, absent, token file present, empty token file)
- [x] New test for `get_content_credentials` returning `None` without token
- [x] Warning logged when token file is empty
- [x] All 398 tests pass

### DIFF
--- a/src/posit/connect/oauth/oauth.py
+++ b/src/posit/connect/oauth/oauth.py
@@ -49,7 +49,7 @@ def _get_content_session_token() -> Optional[str]:
         except FileNotFoundError:
             pass
 
-    return os.environ.get("CONNECT_CONTENT_SESSION_TOKEN") or None
+    return os.environ.get("CONNECT_CONTENT_SESSION_TOKEN")
 
 
 class OAuth(Resources):
--- a/src/posit/connect/oauth/oauth.py
+++ b/src/posit/connect/oauth/oauth.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 
 from typing_extensions import TYPE_CHECKING, Optional, TypedDict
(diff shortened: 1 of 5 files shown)


src/posit/connect/oauth/oauth.py
```diff
@@ -19,40 +20,36 @@
...
         except FileNotFoundError:
             pass
 
-    value = os.environ.get("CONNECT_CONTENT_SESSION_TOKEN")
-    if not value:
-        raise ValueError(
-            "Invalid value for 'CONNECT_CONTENT_SESSION_TOKEN': Must be a non-empty string."
-        )
-    return value
+    return os.environ.get("CONNECT_CONTENT_SESSION_TOKEN") or None
```
nealrichardson: Doesn't `.get()` return `None` if not found anyway?

```suggestion
    return os.environ.get("CONNECT_CONTENT_SESSION_TOKEN")
```

nealrichardson (approved):
One suggestion but I like the improved user experience here!

PR #473, opened 2026-05-07
fix: Bundle.download() reads 1 byte at a time

jimhester:
Bundle.download() called response.iter_content() without chunk_size, which defaults to 1 in requests. For large bundles this makes downloads CPU-bound at ~20 MB/min and looks like a hang.

Pass chunk_size=65536. Measured: 657 MB bundle goes from ~30 minutes (extrapolated) to 33 seconds.

### DIFF
--- a/src/posit/connect/bundles.py
+++ b/src/posit/connect/bundles.py
@@ -14,6 +14,11 @@ if TYPE_CHECKING:
     from .context import Context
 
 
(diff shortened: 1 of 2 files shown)


jimhester:
Fixes #474

tdstein (approved):
Thank you! 

Please ignore the integration test failures. I have tested those on internal systems and everything checks out.

I will go ahead and merge and make a patch release.

PR #485, opened 2026-07-08
feat: support scheduling content

### LINKED ISSUE #484: Support creating/setting content schedules programmatically
## Summary

Add SDK support for creating and updating the deployment/execution schedule of a content item (scheduled re-render of rendered content, e.g. Jupyter/Quarto ETL notebooks).

## Motivation

Users building Python-based CI/CD pipelines want to manage schedules **as code** rather than clicking through the Connect UI. Today there is no supported, versioned Connect API for writing schedules, and no SDK surface for it:

- The versioned Connect API is GET-only for schedules; the content `PATCH` has no writable recurrence field.

Exposing this through the SDK gives users a stable interface now, and insulates them from the underlying API changing later — when a properly documented/versioned endpoint lands, the library implementation swaps out without breaking user code.

## Proposed interface (sketch)

```python
client = connect.Client()
content = client.content.get("<guid>")

# read
content.schedule            # current schedule, or None

# create / set
content.schedule.set(
    kind="dayofweek",       # minute, hour, day, week, dayofweek, dayofmonth, semimonth, year, ...
    days=["monday", "wednesday"],
    start_time="2026-01-01T09:00:00Z",
    timezone="America/New_York",
)

# remove
content.schedule.delete()
```

Exact shape TBD — should mirror the Connect schedule model and align with the R `connectapi` semantics.

## Notes / open questions

- Backing this on the unversioned `POST /__api__/schedules` endpoint initially (with a clear internal note that it's unversioned) so the public SDK method is stable regardless of future API work.
- Related: #189 (Variants API — render + email).

karawoo:
This adds a way to manage schedules for content on Connect. Closes #484.

The default variant's schedule can be managed with `content.schedule`, which supports looking up the schedule (`find_one()`), creating/updating a schedule (`set()`), and removing a schedule (`delete()`). `variant.schedules` does the same for a specific variant. `set()` creates or updates as needed and has per-type overloads, so pyright catches invalid parameter combinations.

Because this is backed by unversioned endpoints, `set()` and `delete()` emit `FutureWarning` similar to how we do for `send_mail`: https://github.com/posit-dev/posit-sdk-py/blob/286d6041a92233ec338bad1f16be38f55a1ee093/src/posit/connect/variants.py#L34-L38

I'm a little on the fence about these warnings and am open to removing them. I do think it's good to convey that this functionality relies on unversioned endpoints that Connect may change or remove.

### DIFF
--- a/integration/tests/posit/connect/test_schedules.py
+++ b/integration/tests/posit/connect/test_schedules.py
@@ -39,7 +39,7 @@ class TestSchedules:
 
         assert schedules.find_one() is None
 
-        created = schedules.set(
+        created = schedules.create(
             type="dayofweek",
             days=["monday", "wednesday"],
             start_time="2026-01-01T09:00:00Z",
@@ -54,7 +54,7 @@ class TestSchedules:
         assert found["id"] == created["id"]
         assert found.rule == {"Days": [1, 3]}
 
-        updated = schedules.set(type="hour", n=2, email=True)
+        updated = schedules.create(type="hour", n=2, email=True)
         assert updated["id"] == created["id"]
         assert updated.rule == {"N": 2}
         assert updated["email"] is True
@@ -90,7 +90,7 @@ class TestSchedules:
         ]
         try:
             for kwargs, expected_rule in cases:
-                result = schedules.set(**kwargs)
+                result = schedules.create(**kwargs)
                 assert result.rule == expected_rule, kwargs
                 found = schedules.find_one()
                 assert found is not None
--- a/integration/tests/posit/connect/test_schedules.py
+++ b/integration/tests/posit/connect/test_schedules.py
@@ -1,0 +1,109 @@
+from pathlib import Path
+
+import pytest
+from packaging import version
+from typing_extensions import Any, Dict, List, Tuple
+
+from posit import connect
+
+from . import CONNECT_VERSION
+
+
+@pytest.mark.skipif(
+    CONNECT_VERSION <= version.parse("2023.01.1"),
+    reason="Quarto not available",
+)
+class TestSchedules:
+    @classmethod
+    def setup_class(cls):
+        cls.client = connect.Client()
+        content = cls.client.content.create(name="example-quarto-minimal")
+
+        path = Path("../../../resources/connect/bundles/example-quarto-minimal/bundle.tar.gz")
+        path = Path(__file__).parent / path
+        bundle = content.bundles.create(str(path.resolve()))
+        task = bundle.deploy()
+        task.wait_for()
+
+        # Re-fetch so the content item reflects the deployed application mode
+        # (the record from `create` predates the deploy and still has app_mode
+        # "unknown", which content.schedule rejects).
+        cls.content = cls.client.content.get(content["guid"])
+
+    @classmethod
+    def teardown_class(cls):
+        cls.content.delete()
+        assert cls.client.content.count() == 0
+
+    def test_lifecycle(self):
+        schedules = self.content.schedule
+
+        assert schedules.find_one() is None
+
+        created = schedules.set(
+            type="dayofweek",
+            days=["monday", "wednesday"],
+            start_time="2026-01-01T09:00:00Z",
+            timezone="America/New_York",
+        )
+        assert created.rule == {"Days": [1, 3]}
+        assert created["activate"] is True
+        assert created["timezone"] == "America/New_York"
+
+        found = schedules.find_one()
+        assert found is not None
+        assert found["id"] == created["id"]
+        assert found.rule == {"Days": [1, 3]}
+
+        updated = schedules.set(type="hour", n=2, email=True)
+        assert updated["id"] == created["id"]
+        assert updated.rule == {"N": 2}
+        assert updated["email"] is True
+        assert updated["activate"] is True
+        # fields not specified in the update are preserved
+        assert updated["timezone"] == "America/New_York"
+        assert updated["start_time"] == "2026-01-01T09:00:00Z"
+
+        schedules.delete()
+        assert schedules.find_one() is None
+
+        # idempotent
+        schedules.delete()
+        assert schedules.find_one() is None
+
+    def test_every_schedule_type(self):
+        schedules = self.content.schedule
+        cases: List[Tuple[Dict[str, Any], dict]] = [
+            ({"type": "minute", "n": 15}, {"N": 15}),
+            ({"type": "hour", "n": 2}, {"N": 2}),
+            ({"type": "day", "n": 3}, {"N": 3}),
+            ({"type": "weekday"}, {}),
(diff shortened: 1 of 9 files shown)


src/posit/connect/content.py
```diff
@@ -675,6 +670,64 @@ def owner(self) -> dict:
...
+
+        # Remove the schedule
+        content.schedule.delete()
+        ```
+        """
+        if not self.is_rendered:
+            raise ValueError(
+                f"Scheduling is not supported for this application mode: {self['app_mode']}. "
+                "Schedules require rendered content, such as R Markdown, Jupyter, or Quarto documents.",
+            )
```
lucasrod16, line 732: `content.schedule` currently rejects when `app_mode` is `unknown`, which is a valid server state immediately after create and before/just after deploy. The integration test works around this by re-fetching. Could we do that refresh here when `app_mode == "unknown"` before raising an error so users don't need to handle this themselves?

https://github.com/posit-dev/posit-sdk-py/pull/485/changes#diff-ca014173ee4bef7d0c24b497d88fae51709e4a5e1b254acec0812b6fcc785b6cR28-R31

lucasrod16 (commented):
Taking a first pass at this and will followup with another pass. The overall API surface looks good to me so far

integration/tests/posit/connect/test_schedules.py
```diff
@@ -0,0 +1,109 @@
...
+                assert result.rule == expected_rule, kwargs
+                found = schedules.find_one()
+                assert found is not None
+                assert found.rule == expected_rule, kwargs
+        finally:
+            if schedules.find_one() is not None:
+                schedules.delete()
+
+    def test_interactive_content_not_schedulable(self):
+        content = self.client.content.create(name="unschedulable")
```
cgraham-rs, line 105: I believe this will result in an "unknown" content type instead of "interactive" since nothing is deployed. It still seems reasonable to assert that content with no bundle cannot be scheduled, but the test name does not match the behavior. I would never have thought of this without the code comments about "unknown" above, and Lucas' PR comment.

src/posit/connect/schedules.py
```diff
@@ -0,0 +1,446 @@
...
+            raise ValueError(f"Invalid n: {n}. Expected an int greater than or equal to 1.")
+        return json.dumps({"N": n})
+    if type == "weekday":
+        return json.dumps({})
+    if type == "dayofweek":
+        if days is None:
+            raise ValueError("Schedule type 'dayofweek' requires 'days'.")
+        return json.dumps({"Days": _normalize_days(days)})
+    if type == "semimonth":
+        return json.dumps({"First": True if first is None else first})
```
lucasrod16, line 131: Can we add a `TypeError` check here to ensure `semimonth.first` is a bool and test cases that cover passing in invalid types?

lucasrod16 (approved):
This appears to be the first API resource in this SDK to use a `set()` method as opposed to `create()`/`update()`, but I think it provides a better UX since users don't have to know about or handle the create vs update decision and details themselves.

I think we should handle the `app_mode == "unknown"` case in `content.schedule` before merge, otherwise this LGTM.

src/posit/connect/variants.py
```diff
@@ -4,10 +4,22 @@
...
 
 class Variant(BaseResource):
+    @property
+    def schedules(self) -> Schedules:
+        """Manager for this variant's render schedule.
+
+        Warnings
+        --------
+        The schedule API is backed by unversioned Connect endpoints and is
+        experimental; it may change in future releases.
```
tdstein: nit; "unversioned" is an implementation detail of Connect. Probably not relevant to end users.

```suggestion
        The schedule API is experimental and may change in future releases.
```

src/posit/connect/schedules.py
```diff
@@ -0,0 +1,446 @@
...
+        type: Literal["dayweekofmonth"],
+        day: int,
+        week: int,
+        n: int = 1,
+        start_time: Union[datetime, str, None] = None,
+        timezone: Optional[str] = None,
+        email: Optional[bool] = None,
+    ) -> Schedule: ...
+
+    def set(
```
tdstein: nit; @lucasrod16 already mentioned this, but other resources use `create` for POST actions. Using `set` is probably fine in this scenario, but `set` is also builtin in Python. I generally try to avoid reusing the builtin values to avoid confusion.

src/posit/connect/schedules.py
```diff
@@ -0,0 +1,446 @@
...
+
+        Warnings
+        --------
+        This operation is backed by an unversioned Connect endpoint and is experimental.
+        """
+        warnings.warn(
+            "destroy() is experimental and may change in future releases.",
+            FutureWarning,
+            stacklevel=2,
+        )
```
tdstein: The WARN messages are a bit spammy: https://github.com/posit-dev/posit-sdk-py/actions/runs/28979813131/job/85995666456?pr=485#step:5:202

An alternative to including a WARN on every method is to do it once in https://github.com/posit-dev/posit-sdk-py/pull/485/changes#diff-ebeb212671b2a7cb98846a4a929d8d29c6ad131825682b2f9d83c72414550674R683 (`content.schedule`)

src/posit/connect/content.py
```diff
@@ -675,6 +670,64 @@ def owner(self) -> dict:
...
+    @property
+    def _default_variant(self) -> Variant:
+        variants = [variant for variant in self._variants.find() if variant["is_default"]]
+        if len(variants) != 1:
+            raise RuntimeError(
+                f"Found {len(variants)} default variants. Expected 1. This is indicative of a corrupted state.",
+            )
+        return variants[0]
+
+    @cached_property
```
tdstein, line 683: Oh interesting. I've never used `cached_property`. Did you include this to avoid recomputing the default variant? Is there any risk that the default variant changes?

tdstein (approved):
Looks good overall!

Although, the amount of logic required to make the Schedules API idiomatic in Python is unfortunate...

integration/tests/posit/connect/test_schedules.py
karawoo: Good catch, renamed to `test_undeployed_content_not_schedulable` and added a comment explaining that app_mode stays "unknown" after the refresh when nothing is deployed.

src/posit/connect/content.py
karawoo: Yes, it's there to avoid re-resolving the default variant. Without caching, `create()` followed by `find_one()` would look it up twice. I don't think the default variant changes so I think this should be safe.

karawoo:
Thanks for the feedback everyone. I opted to rename from `set` to `create` for consistency, though I do think `set()` has some advantages too.

src/posit/connect/schedules.py
karawoo: Renamed to `create` 👍

PR #486, opened 2026-07-28
test: run each integration test class against a fresh Connect container

### LINKED ISSUE #460: Integration tests fail due to stale state on shared Connect instance
## Problem

Integration tests are failing intermittently due to leftover state on the shared Connect instance. Observed in [CI run #23011094234](https://github.com/posit-dev/posit-sdk-py/actions/runs/23011094234/job/66821545249):

- **2 failures, 12 errors** in integration tests
- **0 failures** in unit tests

### Errors (12)

All `test_groups` tests error with:
```
posit.connect.errors.ClientError: {"error_code": 15, "error_message": "The requested group name is already in use.", "http_status": 409, "http_message": "Conflict"}
```

Several other tests (`test_jobs`, `test_tags`, `test_users`, `test_vanities`) error with `assert 1 == 0` on `content.count()`, indicating content from previous test runs was not cleaned up.

### Failures (2)

Also related to stale content state — tests assume `content.count() == 0` but find pre-existing content items.

## Root cause

Integration test setup/teardown does not reliably clean up resources (groups, content items) on the Connect instance. When a previous test run fails partway through or teardown is incomplete, subsequent runs encounter conflicting state.

## Suggested fix

- Ensure test fixtures clean up created resources in teardown (even on failure), or
- Use unique names per test run (e.g., with a timestamp/UUID suffix) to avoid conflicts, or
- Add a pre-test cleanup step that removes resources from prior runs

cgraham-rs:
Fixes #460

- Provisions a fresh, disposable Connect container per test class via a class-scoped autouse fixture wrapping `with-connect` so state can never leak between test classes
- Cleans up the default Connect OAuth integration so related tests have a deterministic baseline and to prevent cross-test pollution. This is done via a scoped `conftest.py` fixture under `oauth/`
- Container lifecycle is now owned by pytest instead of the Makefile
- Updates CI to write the license file directly and drop the composite `with-connect` action
- Documents the new integration test workflow in `integration/README.md`

### DIFF
--- a/integration/tests/posit/connect/conftest.py
+++ b/integration/tests/posit/connect/conftest.py
@@ -1,0 +1,101 @@
+"""Provision a fresh Posit Connect container for each test class.
+
+Prevents state from leaking between classes (posit-dev/posit-sdk-py#460).
+Standard pytest selection works: running a single test boots exactly the
+container(s) its class needs.
+"""
+
+from __future__ import annotations
+
+import os
+import socket
+import subprocess
+import warnings
+
+import pytest
+
+# with-connect requires Python >= 3.13; pin the source to avoid a stale
+# copy on PATH.
+_WITH_CONNECT = [
+    "uvx",
+    "--python",
+    "3.13",
+    "--from",
+    "git+https://github.com/posit-dev/with-connect.git",
+    "with-connect",
+]
+
+# Posit Connect license file to mount into each container (gitignored).
+_LICENSE = os.environ.get("LICENSE", "./license.lic")
+
+
+def _free_port() -> int:
+    with socket.socket() as s:
+        s.bind(("", 0))
+        return s.getsockname()[1]
+
+
+@pytest.fixture(scope="class", autouse=True)
+def fresh_connect():
+    """Boot a fresh Connect container and expose it via the environment.
+
+    Setup runs before setup_class; the finalizer runs after teardown_class,
+    even if it raises.
+    """
+    version = os.environ["CONNECT_VERSION"]
+    port = _free_port()
+
+    result = subprocess.run(
+        [
+            *_WITH_CONNECT,
+            "--version",
+            version,
+            "--port",
+            str(port),
+            "--license",
+            _LICENSE,
+            "--quiet",
+        ],
+        capture_output=True,
+        text=True,
+        timeout=900,  # the first invocation may pull the Connect image
+    )
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"with-connect failed to start Connect {version} on port {port}:\n{result.stderr}"
+        )
+
+    creds = dict(line.split("=", 1) for line in result.stdout.splitlines() if "=" in line)
+
+    # Fetch container_id before validating credentials, so a container is still
+    # cleaned up in `finally` even if with-connect's output is incomplete.
+    container_id = creds.get("CONTAINER_ID")
+    try:
+        missing = [
+            key
+            for key in ("CONNECT_SERVER", "CONNECT_API_KEY", "CONTAINER_ID")
+            if key not in creds
+        ]
+        if missing:
+            raise RuntimeError(
+                f"with-connect start-only output was missing expected credentials: {', '.join(missing)}"
+            )
+
+        os.environ["CONNECT_SERVER"] = creds["CONNECT_SERVER"]
+        os.environ["CONNECT_API_KEY"] = creds["CONNECT_API_KEY"]
+        yield
+    finally:
+        # `rm` (not `stop`) so per-class containers don't accumulate on disk.
+        if container_id is not None:
+            stop = subprocess.run(
+                ["docker", "rm", "--force", "--volumes", container_id],
+                capture_output=True,
+                text=True,
+                timeout=120,
+            )
+            if stop.returncode != 0:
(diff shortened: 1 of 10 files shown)


cgraham-rs:
@tdstein Beyond looking at the PR changes here I also hit a lint error in CI that was not evident locally. It looks like CI pulled a different version of `ruff` which expanded the ruleset.  Things to discuss or create issues for.
- Because `uv.lock` is gitignored, CI resolves dev dependencies fresh on every run while local environments stay pinned to whenever the lock was first created and the two will drift over time. I was running an older `ruff` which passes lint, CI pulled a newer version which is failing lint.
- There's no current make target or CONTRIBUTING instruction to help users update local dev dependencies.
- The latest `ruff` appears to include a variety of new rules which are throwing a significant amount of new lint errors across the entire repo. This may be related to using `extend-select`

integration/tests/posit/connect/conftest.py
tdstein, line 1: Gah! This is so cool. Sometime I just absolutely love Python!

integration/tests/posit/connect/conftest.py
```diff
@@ -0,0 +1,101 @@
...
+            if key not in creds
+        ]
+        if missing:
+            raise RuntimeError(
+                f"with-connect start-only output was missing expected credentials: {', '.join(missing)}"
+            )
+
+        os.environ["CONNECT_SERVER"] = creds["CONNECT_SERVER"]
+        os.environ["CONNECT_API_KEY"] = creds["CONNECT_API_KEY"]
+        yield
```
tdstein, line 86: Beautiful! ✨

tdstein (approved):
This is so great. Thank you so much for putting in the effort to make this happen. What are your thoughts on utilizing this pattern more for doing end-to-end integration testing as part of the Connect CI workflow. That's always been part of the mission here, but it hasn't materialized.

Also, I know there is a bunch of logic in the integration tests that does manually cleanups to make things work between tests. I can delete all of that now, right?

tdstein:
> @tdstein Beyond looking at the PR changes here I also hit a lint error in CI that was not evident locally. It looks like CI pulled a different version of `ruff` which expanded the ruleset. Things to discuss or create issues for.
> 
> * Because `uv.lock` is gitignored, CI resolves dev dependencies fresh on every run while local environments stay pinned to whenever the lock was first created and the two will drift over time. I was running an older `ruff` which passes lint, CI pulled a newer version which is failing lint.
> * There's no current make target or CONTRIBUTING instruction to help users update local dev dependencies.
> * The latest `ruff` appears to include a variety of new rules which are throwing a significant amount of new lint errors across the entire repo. This may be related to using `extend-select`

Yeah, that keeps causing headaches. Is the fix to simply commit `uv.lock`? I don't recall why we ignored it to begin with.

cgraham-rs:
> > @tdstein Beyond looking at the PR changes here I also hit a lint error in CI that was not evident locally. It looks like CI pulled a different version of `ruff` which expanded the ruleset. Things to discuss or create issues for.
> > 
> > * Because `uv.lock` is gitignored, CI resolves dev dependencies fresh on every run while local environments stay pinned to whenever the lock was first created and the two will drift over time. I was running an older `ruff` which passes lint, CI pulled a newer version which is failing lint.
> > * There's no current make target or CONTRIBUTING instruction to help users update local dev dependencies.
> > * The latest `ruff` appears to include a variety of new rules which are throwing a significant amount of new lint errors across the entire repo. This may be related to using `extend-select`
> 
> Yeah, that keeps causing headaches. Is the fix to simply commit `uv.lock`? I don't recall why we ignored it to begin with.

I *think* that is the correct fix. But the fact it's been gitignored for so long gives me pause as there must have been a reason to do that. But that reason was not clear. I created two new issues so we can track and work on those separately.

- #487
- #488

cgraham-rs:
> This is so great. Thank you so much for putting in the effort to make this happen. What are your thoughts on utilizing this pattern more for doing end-to-end integration testing as part of the Connect CI workflow. That's always been part of the mission here, but it hasn't materialized.

I'm not sure what you mean.

> Also, I know there is a bunch of logic in the integration tests that does manually cleanups to make things work between tests. I can delete all of that now, right?

The fresh Connect change is scoped to test classes, not individual tests. The tests within the same class still share one container, so one test's leftover state can still affect a sibling test in that same class. So we will still have to manage that, like we did with the oauth tests in this PR.

cgraham-rs:
IMO we should merge this even though there are lint failures. Those are external to the changes here and are tracked separately.

integration/tests/posit/connect/conftest.py
tdstein: Thanks! The class-scoped fixture is doing the heavy lifting here: each integration test class gets an isolated Connect container, which is cleaned up afterward.

integration/tests/posit/connect/conftest.py
tdstein: Thanks! Keeping the container lifecycle and cleanup in the fixture makes the isolation behavior explicit and reusable across the integration suite.

integration/tests/posit/connect/conftest.py
tdstein: Lol, my bot is talking to itself

PR #492, opened 2026-09-04
fix: remediate Dependabot security alerts

tdstein:
## Summary

- Upgrade all dependencies covered by the 36 open Dependabot alerts to patched versions.
- Raise the supported Python version to 3.10+ and align CI and documentation.
- Update the Flask integration fixture and archived bundle to Flask 3.1.3.
- Fix the Ruff B905 violation surfaced by the refreshed toolchain.

## Validation

- `make test` — 463 passed
- `make lint` — Ruff and Pyright passed
- `make build` — passed
- `uv lock --check` — passed

The patched dependency set requires Python 3.10+, so Python 3.8 and 3.9 are removed from the CI matrix.


tdstein:
@nealrichardson, FYI this changes the minimum supported version of Python from 3.8 to 3.10, but also adds support for 3.14 and 3.15. This matches the current LTS window https://devguide.python.org/versions/. Dropping support for 3.8 and 3.9 is required to resolve the security alerts.

tdstein:
> @nealrichardson, FYI this changes the minimum supported version of Python from 3.8 to 3.10, but also adds support for 3.14 and 3.15. This matches the current LTS window https://devguide.python.org/versions/. Dropping support for 3.8 and 3.9 is required to resolve the security alerts.

I (Luna) also cross-checked with Connect to verify that this doesn't effect any currently published guidance:

```
  - The installation recipe at docs/cookbook/getting-started/installing-the-posit-sdk/index.qmd:17-27 says pip install posit-sdk but does not state the new Python 3.10+ prerequisite. It should gain an explicit note.
  - No Connect documentation mentions the SDK’s CI matrix, so Python 3.14/3.15 requires no documentation update there.
  - References to Python 3.8 in Connect’s admin/content docs describe Connect runtime support, not SDK support; they should remain unchanged.
  - Cookbook examples pin historical SDK versions (0.2.2, 0.5.0, 0.8.0, 0.9.0, 0.12.0). These don’t become invalid from this change, but they are independently stale relative to the Cookbook’s goal of using current
    SDK versions.
```

nealrichardson (approved):

